PR #23, opened 2014-01-04
WIP: Read undocumented datatypes

mbauman:
I've begun work on reverse-engineering some of Matlab's undocumented datatypes — most prominently their new-style class objects.  My group works very extensively with class-based objects within Matlab, and so I've always found that to be a roadblock in picking up, e.g., Octave or NumPy or Julia (yes, yes, I know I can just save them as structs, but that adds additional complexities on the Matlab side that I'd prefer not do deal with).  Plus, this has been a very interesting challenge.

I'm making good progress (just within v5 files), but I'm at the point where I'll need to do some major refactoring; all of the composite datatype readers (e.g., cells, structs, etc) will need knowledge of the information contained within the subsystem.  I think it'd make sense to make them all functions of the Matlabv5File (which would then additionally contain the parsed subsystem information).  I figured I'd ping you before making major changes… does this seem reasonable to you? Or do you have other suggestions?


jebej:
I have adapted a the code in here (and the IJulia notebook linked) to work with v7.3 MAT files. My purpose was to read `datetime` values, which works fine. Should I make an other PR?

There are also various discussion items that have not been resolved e.g.: should the `MCOS` variable be read when the file is opened, so that references to the `#subsystem#` need only to be processed once?

jebej:
BTW that was some epic reverse engineering @mbauman !

jjyyxx:
@mbauman

Your notebook has been extremely helpful for me to implement a v7.3 parser in python, supporting a range of built-in class-like types and custom classes.

However, one thing I feel confused is that:

```julia
if flag == 0
    # This means that the property is stored in the names array
    d[names[name_idx]] = names[heap_idx]
elseif flag == 1
    # The property is stored in the MCOS FileWrapper__ heap
    d[names[name_idx]] = heap[heap_idx+3] # But... the index is off by 3!? Crazy.
elseif flag == 2
    # The property is a boolean, and the heap_idx itself is the value
    @assert 0 <= heap_idx <= 1 "boolean flag has a value other than 0 or 1"
    d[names[name_idx]] = bool(heap_idx)
else
    error("unknown flag ",flag, " for property ",names[name_idx], " with heap index ",heap_idx)
end
```

The comment "**The property is a boolean, and the heap_idx itself is the value**" makes sense to me at first, before I meet a counterexample with `heap_idx` equal to `-1`.

I'm seeking help because I was unable to write a custom class example with fields saved with flag `2`. That is, I can't determine the exact condition for flag 2. I can only reproduce by creating instances of `RTW.TimingInterface` (field `Priority`) or `coder.types.AggregateElement` (field `TargetOffset`). However, they are built-in classes that are hard to inspect.

**Things I've tried so far:**
1. Get metaclass of `RTW.TimingInterface`, and inspect its properties. I made my own class align with `RTW.TimingInterface`, but still could not reproduce this behavior.
2. I noticed that fields mentioned above, for example, `RTW.TimingInterface#Priority` enforces scalar input. In my custom class, add property validation (size specification), hinting the property is a scalar, but still could not reproduce this behavior.

I would like to know how you find flag 2 and its meaning. Could you provide a sample class definition or some assistance?

Sorry for asking under this PR if this is inappropriate.

Jeroen-van-der-Meer:
I know this is an ancient PR, but I'm wondering if people are still interested in supporting this in MAT.jl.

I have reverse-engineered how MATLAB stores classes in the MAT v7.3 format. I am planning to write up my findings in a couple of posts on my homepage. So far two have been written up:

1. In [the first post](https://meer.gitlab.io/blog-7.html) I've just given a general review of how MATLAB stores simple objects in v7.3. None of this is new, and you're familiar with the MAT.jl code then you can safely ignore this.
2. In [the second post](https://meer.gitlab.io/blog-8.html) I've outlined how MATLAB stores classes. The hardest part is figuring out how MATLAB stores the 'tree structure' when you have classes within classes.

There are a few more things that I'll write up as soon as I find the time.

* If your class has default values or it has properties that have their types specified, the encoding changes, and I'll highlight how to deal with that.
* Once you understand classes, it's pretty easy to parse datetimes and tables, and I'll elaborate a bit on how that's done.
* Strings and enums are encoded in a rather peculiar way, not consistent with other objects, and they have to be dealt with separately.

@mbauman I'd be interested to know to what extent my findings will match up with what you have studied. I have not looked at MAT v5, and I expect that there'll be a lot of similarities.

ViralBShah:
@Jeroen-van-der-Meer Happy to give you access to this repo if you'd like to improve the package.

foreverallama:
Hi! I was recently made aware of this PR. I've been working on something [very similar](https://github.com/foreverallama/matio) in Python (built around SciPy). Been able to (almost) completely reverse engineer MAT-files here, as well as build conversion utilities between MATLAB and Pythonic datatypes.

If this still has sufficient interest, I wouldn't mind porting it over here as well. I haven't worked with Julia before so it would be an interesting challenge. Let me know!

ViralBShah:
Contributions always welcome! We may not have active maintenance here, but this would be a good way to participate.

foreverallama:
Good to hear that, thanks! Just a question before try anything out, is there any preferred way to represent class objects. I see in the current code for `mxOBJECT_CLASS` you're writing a `Dict` with an extra key "class" for the class name. Would that work here as well - a dictionary of property name, property value, and an extra key "class" for the class name?

foreverallama:
I opened a new PR related to this in #205, will be adding in parts that'll make it easier to review. Any feedback is appreciated as I'm new to writing in Julia

matthijscox:
Thanks to @foreverallama (and me) we can finally close this old PR :). 

MAT v0.11 will have official support for the `MatlabClassObject` (old class) and the new `MatlabOpaque` (new class) types.

mbauman:
Amazing! Bravo to @foreverallama; I'm thrilled to see this finally better supported!

foreverallama:
All thanks to @matthijscox for coming up with crucial updates (and a lot of enthusiasm) to help push this through! 🙌

mbauman:
Also as an aside, if anyone here still has an active Matlab license, it'd probably be good to check if my MWE in https://github.com/JuliaIO/MAT.jl/pull/23#issuecomment-37745085 still segfaults on the latest release.  If it does, it'd be very good to report it to The MathWorks; it's a rather serious security issue.

foreverallama:
It still does. As far as I understand, its definitely intended behaviour. The gist is, any `uint32` vector that starts with `0xDD000000` is a placeholder for objects. MATLAB assigns a unique ID (basically a counter going up from one) to each object instance saved in a MAT-file. This `uint32` vector contains metadata like object dimensions, the object IDs and the class ID. In your example, MATLAB would read the property value as an object instance with an `ID=2`, which doesn't exist in the file (there is only one object instance - the container.Map) and leads to a segfault. 

In fact, you can segfault in many number of ways. I suspect almost all of them are from going out of bounds in array or vector operations. It's a bit surprising that they haven't introduced any safety checks,, but in any case, it's definitely a design choice as the subsystem codebase has definitely gone through several version updates over the years.

If you're interested, I've [documented](https://github.com/foreverallama/matio/blob/main/docs/subsystem_data_format.md) the subsystem format in detail. My theory is that the whole idea of the subsystem came about from their `.fig` format. But as a bonus, there doesn't seem to be any difference between `.fig` and `.mat` formats, and the latest update to `MAT.jl` should be able to read figures as well. This in fact opens up more opportunities, for example saving figures in MATLAB and loading them straight in Python/Octave/Julia or vice versa (which I'm working on).

PR #31, opened 2014-08-16
Support for [] getindex/setindex! syntax. Fix #30

### LINKED ISSUE #30: `file["varname"]` and `file[:varname]` interface?
I just found myself intuitively expecting that, probably because I'm used to it from scipy io and jsoncpp.

Would you want a PR for that, or is it too non-Julian? (Or is there a caveat which I'm not seeing?)

lucasb-eyer:
A pretty simple change, but I still need comments:
1. Why do the `write` methods get a `ByteString` for the variable name while the `read` methods get an `ASCIIString`?
2. Do you want the `Symbol` syntax `file[:varname]`, or is that too many ways to do the same thing?

### DIFF
--- a/src/MAT_HDF5.jl
+++ b/src/MAT_HDF5.jl
@@ -28,7 +28,7 @@
 
 module MAT_HDF5
 using HDF5
-import Base: read, write, close
+import Base: read, write, close, getindex, setindex!
 import HDF5: names, exists, HDF5ReferenceObj, HDF5BitsKind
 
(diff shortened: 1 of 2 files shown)


matthijscox:
Hi, new MAT.jl maintainer here. While this is a fun and simple feature, it's a really old PR, so I'm considering to close it.

lucasb-eyer:
haha nice throwback :smile:
definitely do as you feel! I didn't remember this PR, and haven't used this code in about a decade.

PR #46, opened 2015-10-09
Fix of deprecated warnings Julia 0.4.0

RobertBeyer:


src/MAT_HDF5.jl
```diff
@@ -278,7 +278,7 @@ function m_writetypeattr(dset, T)
 end
 
 # Writes an empty scalar or array
-function m_writeempty(parent::Union(HDF5File, HDF5Group), name::ByteString, data::Array)
```
simonster: This and other changes from `Union(...)` to `Union{...}` need to be `@compat(Union{...})` to avoid breaking on 0.3.

simonster:
Closed by #48

PR #48, opened 2015-10-12
Update for latest syntax

phobon:
Hello, 

I've updated the code to fix deprecation warnings about:
- `String` instead of `AbstractString`
- `None` instead of `Union{}`
- `Union(args...)` instead of Union{args...}`

Now the module import and tests run without warning for me on `0.5`. I also made sure `Pkg.test("MAT")` still works fine on `0.3`.

It would be great if we can get a tagged version after this to coincide with the `0.4` release.

Thanks,
Josh

### DIFF
--- a/src/MAT.jl
+++ b/src/MAT.jl
@@ -32,8 +32,8 @@ using HDF5, MAT_HDF5, MAT_v5
 export matopen, matread, matwrite, names, exists, @read, @write
 
 # Open a MATLAB file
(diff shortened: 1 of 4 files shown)


simonster:
Thanks! I will update Travis to test on 0.3 and then tag this.

simonster:
Tagged. Thanks again!

phobon:
Thanks for the quick response!

PR #54, opened 2016-04-14
added docstrings

bjarthur:
nice to have these on the REPL so i don't have to reference the README.

### DIFF
--- a/src/MAT.jl
+++ b/src/MAT.jl
@@ -108,7 +108,24 @@ function matopen(f::Function, args...)
     end
 end
 
+"""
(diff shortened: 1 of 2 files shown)


timholy:
I'd like this. Looks like it could use a rebase.

bjarthur:
thanks tim!  tests now pass on all but nightly.

has moving this MAT.jl to JuliaIO been considered?

timholy:
I thought it had perhaps been discussed, but a brief search didn't turn up the issue for me.

PR #57, opened 2016-05-20
fix reading of mxOBJECT_CLASS

tanmaykm:
reading of class name was misplaced, resulting in corruption/errors.


tanmaykm:
cc @andreasnoack

timholy:
Would benefit from a test, if you can create a relevant file.

tanmaykm:
This file here https://www.dropbox.com/s/6b2hi253otmkonj/3255538_0001.mat?dl=0 displays the issue. I think it is too big to be included for tests. It can be trimmed (to just `A` that's in it), but I do not have access to matlab. It would be great if someone can trim it or make an equivalent one.

timholy:
When I try to read it in matlab, I get:
```
Warning: Class 'Assoc' is an unknown object class or does not have a valid 'loadobj' method.  Object 'A' of this class has been converted to a structure.
```
So I can't trim it down either, the original author will have to do it.

andreasnoack:
Thanks for looking into this @tanmaykm. I didn't notice the warning when I loaded it in Matlab. I looked at the header and it seems that the file was generated in Octave. Maybe Octave are doing something wrong here?

This file is quite small and has the same issue so maybe we can use it for the test
https://www.dropbox.com/s/xrcl5idzvks1dh3/3700696_0079.mat?dl=0

tanmaykm:
Thanks! Have added it to the tests.

timholy:
> Maybe Octave are doing something wrong here?

Just to check, was the new file generated in Matlab or Octave?

andreasnoack:
Both are created in Octave. They give the same warning when loaded in Matlab, the same error in `MAT.jl` and no error in Octave.

timholy:
Doesn't that suggest that Octave is the one that's broken? Isn't there some risk that you'll break Matlab by making this change?

andreasnoack:
@tanmaykm What are your thoughts? Matlab is able to read the file, but with a warning which seems better that erroring.

simonster:
I checked https://maxwell.ict.griffith.edu.au/spl/matlab-page/matfile_format.pdf and @tanmaykm's correction looks right. The warning in MATLAB doesn't seem surprising since you don't have the .m file that defines the object class.

tanmaykm:
I had the same reasoning as @simonster and the changes seem correct as per
the format spec.

timholy:
:+1:

simonster:
Thanks!

PR #60, opened 2016-08-05
Fix depwarn on 0.5 and 0.6

yuyichao:


yuyichao:
Hmmm, test passes for me locally on 0.6-dev with HDF5.jl master. Need new tag from HDF5?

yuyichao:
So the new HDF5 tag fixes the issue on 0.5 =)

simonster:
Thanks!

PR #64, opened 2017-01-19
Fix depwarns and most of errors on 0.6

yuyichao:
The only error left on 0.6 is https://github.com/JuliaLang/julia/issues/20078

### DIFF
--- a/src/MAT_v5.jl
+++ b/src/MAT_v5.jl
@@ -31,6 +31,14 @@ using Compat.String
 import Base: read, write, close
 import HDF5: names, exists
 
+if VERSION < v"0.6.0-dev.1632"
+    round_uint8(data) = round(UInt8, data)
+    complex_array(a, b) = complex(a, b)
+else
+    round_uint8(data) = round.(UInt8, data)
+    complex_array(a, b) = complex.(a, b)
+end
+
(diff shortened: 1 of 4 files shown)


simonster:
Thanks. LGTM provided tests pass.

src/MAT_v5.jl
```diff
@@ -31,6 +31,14 @@ using Compat.String
 import Base: read, write, close
 import HDF5: names, exists
 
+if VERSION < v"0.6.0-dev.1632"
+    round_uint8(data) = round(UInt8, data)
+    complex_array(a, b) = complex(a, b)
+else
+    round_uint8(data) = round.(UInt8, data)
+    complex_array(a, b) = complex.(a, b)
```
yuyichao: And just a note for future reviewers. These are to workaround non-numbers not treated as scalars for broadcast on 0.5 and the bad type inference on 0.4. The version check is not a exact version for any of these to work but should be new enough for these to work and old enough for these to not raise a depwarn.

yuyichao:
The errors on master are due to JuliaLang/julia#20078 and the depwarns are mostly(all?) from HDF5, which should already be fixed but not tagged yet.

PR #67, opened 2017-05-21
Fix for UInt8 error in close()

laurentheirendt:
Fix for `UInt8` error when using `close()`:
```
ERROR: LoadError: LoadError: type String has no field data
Use `Vector{UInt8}(str)` instead.
Stacktrace:
 [1] close(::MAT.MAT_HDF5.MatlabHDF5File) at /home/vagrant/.julia/v0.6/MAT/src/MAT_HDF5.jl:61
```

cc: @yuyichao, @simonster, @syarra

### DIFF
--- a/src/MAT_HDF5.jl
+++ b/src/MAT_HDF5.jl
@@ -58,7 +58,7 @@ function close(f::MatlabHDF5File)
         if f.writeheader
             magic = zeros(UInt8, 512)
             identifier = "MATLAB 7.3 MAT-file" # minimal but sufficient
-            magic[1:length(identifier)] = Vector{UInt8}(identifier)
+            magic[1:length(identifier)] = Vector{UInt8}("$identifier")
             magic[126] = 0x02
             magic[127] = 0x49
             magic[128] = 0x4d


laurentheirendt:
Julia `0.6` and latest commit on the `master` branch of `MAT.jl` lead to a failure when closing the `.mat` file. 

The log is [here](http://pkg.julialang.org/logs/COBRA_0.6.log).

yuyichao:
No it's not the latest commit on master. This just need a tag.

laurentheirendt:
In the `.travis` file of the `COBRA.jl` [repository](https://github.com/opencobra/COBRA.jl/blob/master/.travis.yml#L38), the latest commit on `master` is checked out, so how come it is not the latest commit?

Actually, I asked for a new tag a while ago in [this issue](https://github.com/simonster/MAT.jl/issues/66)...

yuyichao:
PkgEval is not travis.

yuyichao:
And FWIW your travis did pass https://travis-ci.org/opencobra/COBRA.jl

laurentheirendt:
Ok, so how can we fix the `PkgEval` ... ?

yuyichao:
A tag will fix it. The last tag wasn't made with attobot so it'll be much easier if an admin on this repo enables it.

PR #77, opened 2017-10-01
Added support for compression

fguevaravas:
MAT.jl already supports compression when reading mat files. This pull request implements **compression in write** mode, that Matlab/Octave supports (MATLAB v7.3 format). Here is a sample usage
```julia
# create a big matrix that compresses well
A = zeros(1000,1000); A[4,5] = 1; A[999,3]=1;

# uncompressed
f=matopen("toto.mat","w"); write(f,"A",A); close(f)

# compressed
f=matopen("totoz.mat","wz"); write(f,"A",A); close(f)

# compare
f=matopen("totoz.mat"); B=read(f,"A"); close(f)
norm(vec(A-B),Inf)
```
Enabling compression reduces size from 8MB to 5k in this (contrived) example.

Some other details:
* Heuristic chunking is used, as in [JLD.jl](https://github.com/JuliaIO/JLD.jl)
* Shuffling is disabled but that can be enabled easily
* The write tests were modified to test both "w" and "wz" flags. 
* Tests pass in Julia 0.6.0 (Mac OS X 10.12.6)


nistath:
I'd be really interested in this getting merged.

timholy:
Incorporated in #123

Issue #84, opened 2018-06-07
Can't write a BitArray to a MAT file


RobBlackwell:
Writing a `BitArray` doesn't work:

````
using MAT

A = trues(10,10)

matwrite("foo.mat", Dict("mask"=>A))

# ERROR: This is the write function for CompositeKind, but the input doesn't fit
# Stacktrace:
#  [1] m_write(::MAT.MAT_HDF5.MatlabHDF5File, ::HDF5.HDF5Group, ::String, ::Tuple{Int64,Int64}) at /home/reb/.julia/v0.6/MAT/src/MAT_HDF5.jl:514
#  [2] m_write(::MAT.MAT_HDF5.MatlabHDF5File, ::HDF5.HDF5File, ::String, ::Array{String,1}, ::Array{Any,1}) at /home/reb/.julia/v0.6/MAT/src/MAT_HDF5.jl:502
#  [3] m_write(::MAT.MAT_HDF5.MatlabHDF5File, ::HDF5.HDF5File, ::String, ::BitArray{2}) at /home/reb/.julia/v0.6/MAT/src/MAT_HDF5.jl:517
#  [4] write(::MAT.MAT_HDF5.MatlabHDF5File, ::String, ::BitArray{2}) at /home/reb/.julia/v0.6/MAT/src/MAT_HDF5.jl:530
#  [5] matwrite(::String, ::Dict{String,BitArray{2}}) at /home/reb/.julia/v0.6/MAT/src/MAT.jl:156
````

A workaround is to convert to an Array of Bool and this becomes a MATLAB logical array

````
matwrite("bar.mat", Dict("mask"=>convert(Array{Bool},A)))
````

It would be nice if the MAT.jl library handled this conversion automatically on write.

PR #85, opened 2018-06-07
Save BitArray as logical array

RobBlackwell:
Fix for https://github.com/JuliaIO/MAT.jl/issues/84


timholy:
Sorry no one noticed this! Because your repo was deleted, I added (with a test :wink:) in #124

PR #91, opened 2018-08-21
tests pass in julia 1.0

denizyuret:
Julia 1.0 compatible version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ julia:
     - nightly
 notifications:
     email: false
-git:
-    depth: 9999999
 addons:
   apt:
     packages:
@@ -16,4 +14,4 @@ addons:
 before_script:
     - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; brew link --overwrite gcc; brew install hdf5; fi
 after_success:
-    - julia -e 'using Pkg, MAT; cd(joinpath(dirname(pathof(MAT)), "..")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder()); Coveralls.submit(process_folder())'
+  - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder()); Codecov.submit(Codecov.process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,11 @@ os:
     - linux
     - osx
 julia:
-    - 0.5
-    - 0.6
+    - 1.0
     - nightly
 notifications:
     email: false
-script:
-    - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-    - julia -e 'Pkg.clone(pwd()); Pkg.build("MAT"); Pkg.test("MAT"; coverage=true)'
+git:
+    depth: 9999999
 after_success:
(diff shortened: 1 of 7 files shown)


andreasnoack:
You'd have to update `.travis.yml` as well. Could you try removing the `script` part and enable testing on 1.0?

denizyuret:
Updated travis.

.travis.yml
```diff
@@ -3,13 +3,11 @@ os:
     - linux
     - osx
 julia:
-    - 0.5
-    - 0.6
+    - 1.0
```
alyst: also 0.7? (as in REQUIRE)

src/MAT.jl
```diff
@@ -143,7 +143,7 @@ end
 Write a dictionary containing variable names as keys and values as values
 to a Matlab file, opening and closing it automatically.
 """
-function matwrite{S, T}(filename::AbstractString, dict::Associative{S, T})
+function matwrite(filename::AbstractString, dict::AbstractDict{S, T}) where {S, T}
```
alyst: IIUC type params unnecessary

.travis.yml
denizyuret: General recommendation was no 0.7 testing if 1.0 passes.

src/MAT.jl
denizyuret: cleaned up.

alyst:
Looks like CI fails during HDF5.jl installation. I'm not sure what's the issue, but HDF5.jl has
```yml
addons:
  apt:
    packages:
    - hdf5-tools
```
in its .travis.yml

denizyuret:
I don't think we get an error with the new HDF5 0.10.  My tests pass.  I updated REQUIRE.

denizyuret:
CI failed again. This seems to be a problem with CI building HDF5 not being able to download stuff. Builds and tests ok on a regular machine.  Can we release this?

alyst:
It looks like Julia tries to install `hdf5-tools` during `Pkg.build()`. IMHO it's still worth to try fixing it by updating travis.yml.

denizyuret:
How do you suggest we do that?

alyst:
Just add `hdf5-tools` as "addon" (see the earlier comment, also [HDF5's travis.yml](https://github.com/JuliaIO/HDF5.jl/blob/master/.travis.yml))

denizyuret:
Seems to fail sporadically.

denizyuret:
And I can't make it fail on my machine to debug :(
Can anybody replicate the CI failure?

quaertym:
@denizyuret It is possible to ssh to the travis CI machine.

denizyuret:
@emreunal Can you tell what is causing the issue?

test/write.jl
```diff
@@ -81,7 +81,7 @@ test_write(Dict(
 @test_throws ErrorException test_write(Dict("another invalid key" => "invalid characters"))
 @test_throws ErrorException test_write(Dict("yetanotherinvalidkeyyetanotherinvalidkeyyetanotherinvalidkeyyetanotherinvalidkey" => "too long"))
 
-type TestCompositeKind
+struct TestCompositeKind
```
woest: Shouldn't `type` be converted to `mutable struct`?

woest: Shouldn't `type` be converted to `mutable struct`?

PR #92, opened 2018-08-24
partial update to 1.0

davidssmith:
doesn't pass all tests, but gets much farther


aliddell:
You should update the .travis.yml to target 0.7 and 1.0 so we don't have these build errors.

PR #95, opened 2018-09-14
Run travis tests on julia >= 0.7 and above as well

alexmorley:


yuyichao:
Included in https://github.com/JuliaIO/MAT.jl/pull/86 and there's little point to turn them on without fixing the support first.

PR #98, opened 2018-10-04
Update for julia 1.0

nicoleepp:
Drops travis 0.5 tests, adds tests for 0.7 and 1.0


yuyichao:
Dup of https://github.com/JuliaIO/MAT.jl/pull/86. Please help with the extra copying issue in that PR instead.

PR #100, opened 2018-10-29
Compatibility for Julia v 1.0.1

gdestouet:
I mostly rewrote :

- some functions definitions with the new syntax ' func(... : : T) where{T}'.

- struct definitions with the keyword 'struct' instead of 'type'

- old variables Vector{UInt8}(nbytes) to Vector{UInt8}(undef,nbytes)


src/MAT.jl
```diff
@@ -143,7 +143,7 @@ end
 Write a dictionary containing variable names as keys and values as values
 to a Matlab file, opening and closing it automatically.
 """
-function matwrite{S, T}(filename::AbstractString, dict::Associative{S, T})
+function matwrite(filename::AbstractString, dict::Dict{S, T}) where{S,T}
```
yuyichao: `AbstractDict`.

src/MAT_HDF5.jl
```diff
@@ -144,14 +145,6 @@ function m_read(dset::HDF5Dataset)
         mattype = a_read(dset, name_type_attr_matlab)
         if mattype == "char"
             return ""
-        elseif mattype == "struct"
```
yuyichao: Don't delete this!

src/MAT_HDF5.jl
```diff
@@ -303,7 +296,7 @@ toarray(x::Bool) = UInt8[x]
 toarray(x) = [x]
 
 # Write the MATLAB type string for dset
-m_writetypeattr{T}(dset, ::Type{Complex{T}}) = m_writetypeattr(dset, T)
+m_writetypeattr(dset, ::Type{Complex{T}}) where{T} = m_writetypeattr(dset, T)
```
yuyichao: EOL whitespace.

src/MAT_HDF5.jl
```diff
@@ -513,7 +506,7 @@ function m_write(mfile::MatlabHDF5File, parent::HDF5Parent, name::String, k::Vec
 end
 
 # Write Associative as a struct
-m_write(mfile::MatlabHDF5File, parent::HDF5Parent, name::String, s::Associative) =
+m_write(mfile::MatlabHDF5File, parent::HDF5Parent, name::String, s::Dict) =
```
yuyichao: `AbstractDict`.

src/MAT.jl
```diff
@@ -53,7 +53,7 @@ function matopen(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Boo
     rawfid = open(filename, "r")
 
     # Check for MAT v4 file
-    magic = read!(rawfid, Vector{UInt8}(4))
+    magic = read!(rawfid, Vector{UInt8}(undef,4))
```
yuyichao: Space after the `,`, same below.

yuyichao:
And this shouldn't drop 0.6 support so you need `Compat.SparseArrays` instead. You'll also need to raise the Compat version required and for 1.0 compatibility changes Compat 1.0 should be good enough.

yuyichao:
Sorry I made a mistake when updating this branch and it appears not revertable so I reopenned this as https://github.com/JuliaIO/MAT.jl/pull/102

PR #106, opened 2018-12-13
Fix some compatibility issues

### LINKED ISSUE #104: Incompatibilities in MAT_v5.jl with v1.0.2
When making my Julia code campatibile with the latest stable version 1.0.2, I have figured out some lines in "MAT_v5.jl" that were still not compatibile, so I have fixed them as follows:

Line 171: `field_name_strings = Vector{String}(undef, n_fields)`
Line 175: `index = findfirst(isequal( 0), sname)`
Line 193: `data[field_name] = Array{Any}(undef, dimensions...)`
Line 260: `data = Vector{String}(undef, dimensions[1])`

With these modifications at least my code works as before.

mforets:
Closes #104.

### DIFF
--- a/src/MAT_HDF5.jl
+++ b/src/MAT_HDF5.jl
@@ -156,7 +156,7 @@ function m_read(dset::HDF5Dataset)
             end
         else
             T = mattype == "canonical empty" ? Union{} : str2eltype_matlab[mattype]
-            return Array{T}(dims...)
+            return Array{T}(undef, dims...)
         end
     end
 
@@ -165,7 +165,7 @@ function m_read(dset::HDF5Dataset)
     if mattype == "cell"
         # Cell arrays, represented as an array of refs
         refs = read(dset, Array{HDF5ReferenceObj})
-        out = Array{Any}(size(refs))
+        out = Array{Any}(undef, size(refs))
         f = file(dset)
         for i = 1:length(refs)
             dset = f[refs[i]]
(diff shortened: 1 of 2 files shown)


yuyichao:
Lgtm as long as 0.6 is still working. (It doesn't seem that this is a complete fix). Should be able to handle with compact if not working already.

mforets:
> Lgtm as long as 0.6 is still working.

The tests is v0.6 pass locally.

> It doesn't seem that this is a complete fix

True. I just fixed some more, but tests in v1.0.2 do not yet pass.
FWIW, the error is:

```julia
$ julia1 --color=yes test/runtests.jl
ERROR: LoadError: LoadError: Data mismatch reading logical_mat from logical.mat (v6)

Got Base.ReinterpretArray{Bool,2,UInt8,Array{UInt8,2}}:

Bool[true false false; false true false; true false false]

Expected Array{Bool,2}:

Bool[true false false; false true false; true false false]
```

I don't plan to send a new commit; I guess it is related to [these lines](https://github.com/JuliaIO/MAT.jl/pull/86/files#diff-ef85ae259948232aba770e48af913465R131) in @halleysfifthinc's PR, but i haven't checked.

yuyichao:
> It doesn't seem that this is a complete fix

That's only relevant for 0.6 compatibility. Not a complete fix isn't a problem at all. No need to put everything in. I can merge as long as CI on 0.6 pass. Feel free to ping. (Not gonna watch it....)

PR #107, opened 2019-01-10
Changes to improve v1.0 handling

dawbarton:
This PR improves compatibility with v1.0 in a similar vein to #86 but uses the low-level HDF5 interface to avoid having to use reinterpret. It doesn't fix all the compatibility problems (I haven't had time) but it shouldn't break v0.6 compatibility either (waiting on CI to confirm).

### DIFF
--- a/src/MAT_v5.jl
+++ b/src/MAT_v5.jl
@@ -137,7 +137,7 @@ end
 function read_data(f::IO, swap_bytes::Bool, ::Type{T}, dimensions::Vector{Int32}) where {T}
     (dtype, nbytes, hbytes) = read_header(f, swap_bytes)
     read_type = READ_TYPES[dtype]
-    if (read_type == UInt8) && (T === Bool)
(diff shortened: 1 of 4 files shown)


dawbarton:
According to CI, this works with v0.6 but (as stated) doesn't fix all the v1.0 issues yet. Hopefully this is a step in the right direction.

src/MAT_v5.jl
```diff
@@ -136,6 +137,9 @@ end
 function read_data(f::IO, swap_bytes::Bool, ::Type{T}, dimensions::Vector{Int32}) where {T}
     (dtype, nbytes, hbytes) = read_header(f, swap_bytes)
     read_type = READ_TYPES[dtype]
+    if (read_type == UInt8) && (T === Bool)
```
yuyichao: `=== UInt8`

yuyichao: Why is this needed now?

PR #118, opened 2019-07-24
Add the ability to load Matlab figure files

dawbarton:
Matlab (including recent versions) appears to save figures as a v6/v7 MAT file. This PR adds in two undocumented classes in order to be able to recognise/open these files. Tests included.


mcmcgrath13:
@dawbarton It looks like this handles more than figures - it also solves the issue of the undocumented classes causing fatal errors.  I've been running into this with audio-related fields and this solves it (in the sense that reading the file doesn't error).  Thanks!

timholy:
Sorry no one noticed, thanks for the contribution!

PR #123, opened 2019-09-20
feat: allow writing compressed files

radioflash:
This fixes the original effort from fguevaravas (#77),
also adding a testcase for compression effectiveness,
correcting some minor whitespace issues, and
adding an example to README.md.

### DIFF
--- a/src/MAT.jl
+++ b/src/MAT.jl
@@ -35,11 +35,11 @@ export matopen, matread, matwrite, names, exists, @read, @write
 
 # Open a MATLAB file
 const HDF5_HEADER = UInt8[0x89, 0x48, 0x44, 0x46, 0x0d, 0x0a, 0x1a, 0x0a]
-function matopen(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Bool, ff::Bool)
+function matopen(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Bool, ff::Bool, compress::Bool)
     # When creating new files, create as HDF5 by default
     fs = filesize(filename)
     if cr && (tr || fs == 0)
-        return MAT_HDF5.matopen(filename, rd, wr, cr, tr, ff)
+        return MAT_HDF5.matopen(filename, rd, wr, cr, tr, ff, compress)
     elseif fs == 0
         error("File \"$filename\" does not exist and create was not specified")
     end
@@ -76,7 +76,7 @@ function matopen(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Boo
         seek(rawfid, offset)
         if read!(rawfid, Vector{UInt8}(undef, 8)) == HDF5_HEADER
             close(rawfid)
-            return MAT_HDF5.matopen(filename, rd, wr, cr, tr, ff)
+            return MAT_HDF5.matopen(filename, rd, wr, cr, tr, ff, compress)
         end
     end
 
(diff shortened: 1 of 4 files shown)


radioflash:
There is still room for improvement:

Reference file (MATLAB 5.0): 5MiB
Julia compressed: 18MiB
Julia uncompressed: 215MiB

Interestingly, the reference file contains no uncompressed strings whatsoever, while the compressed Julia version keeps basically every fieldname as plaintext (my example stores a single Dict: String => Array{Float64, 2}).

But this is currently good enough for me, and it already helps a TON.

src/MAT.jl
```diff
@@ -85,9 +85,10 @@ function matopen(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Boo
 end
 
 function matopen(fname::AbstractString, mode::AbstractString)
-    mode == "r"  ? matopen(fname, true , false, false, false, false) :
-    mode == "r+" ? matopen(fname, true , true , false, false, false) :
-    mode == "w"  ? matopen(fname, false, true , true , true , false) :
+    mode == "r"  ? matopen(fname, true , false, false, false, false, false) :
+    mode == "r+" ? matopen(fname, true , true , false, false, false, false) :
+    mode == "w"  ? matopen(fname, false, true , true , true , false, false) :
+    mode == "wz" ? matopen(fname, false, true , true , true , false, true ) :
```
timholy: Is this standard? Should be documented in any case.

Seems like `z` might be relevant for `r+`? If so, you could check `compress = endswith(mode, 'z')` and if true strip it off and then pass `compress` as the third argument.

src/MAT_HDF5.jl
```diff
@@ -365,7 +374,7 @@ function m_write(mfile::MatlabHDF5File, parent::HDF5Parent, name::String, data::
         m_writeempty(parent, name, data)
         return
     end
-    dset = m_writearray(parent, name, toarray(data))
+    dset = m_writearray(parent, name, toarray(data),mfile.compress)
```
timholy:
```suggestion
    dset = m_writearray(parent, name, toarray(data), mfile.compress)
```

src/MAT_HDF5.jl
```diff
@@ -101,7 +102,7 @@ function matopen(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Boo
         writeheader = false
     end
     close(pa)
-    fid = MatlabHDF5File(HDF5File(f, filename), true, writeheader, 0)
+    fid = MatlabHDF5File(HDF5File(f, filename), true, writeheader, 0,compress)
```
timholy:
```suggestion
    fid = MatlabHDF5File(HDF5File(f, filename), true, writeheader, 0, compress)
```

src/MAT.jl
radioflash: > Is this standard?
I took this from fguevaravas original PR tbh.

Maybe a keyword argument like `matopen(...; compress = true)` both for matopen and matwrite would be best?

Would you prefer keyword argument or "z" options for the mode-string?

timholy: If the `wz` isn't standard, then I think the keyword might be best.

timholy:
Looks like there's a merge conflict, probably from me merging the bit array test. Is your git-fu up to resolving it? If not, don't worry about it, I'll fix it.

src/MAT.jl
radioflash: On it!

timholy:
Thanks @radioflash and @fguevaravas!

PR #125, opened 2019-09-21
Switch to Project.toml

timholy:
Preparation for release 0.5.1


timholy:
I'll release this and then compression can be v0.6

PR #131, opened 2019-12-02
Add upper bounds on all dependencies

timholy:
CC @sciencesam


timholy:
I don't understand the persistent failure for linux 0.7; on my machine (with the same libhdf5 version) it works fine.

PR #132, opened 2019-12-12
v4 support

vsaase:
Hi,
this provides v4 support. I am not yet very experienced with Julia so it might be a bit rough.
Comments welcome.


musm:
Is this good to go?

vsaase:
I think so. The tests with the set of MAT4 files, which use all functionality of MAT4, passed.

musm:
This is causing some test failures I think on Julia 1.5.1 on MacOS for some reason: https://travis-ci.org/github/JuliaIO/MAT.jl/jobs/724195833

I'm not sure why exactly this is Mac specific.

musm:
Actually that error doesn't look related.

musm:
I guess the only thing I'm not a big fan of is naming it `matwrite4` 

can  we not make it kwarg so we can consolidate them all.

sjkelly:
Supplanted by #186. Thanks for getting the ball rolling here!

Issue #136, opened 2020-03-09
How to solve unsatisfiable requirements detected for package error, that occurs while adding MAT?


efmanu:
Hi ,
When I try to add MAT to my packages, I am getting the following error. I can't update `DistributionsAD` because this version required for my application. So, is there any other method to solve this?
ERROR: Unsatisfiable requirements detected for package BufferedStreams [e1450e63]:
 ```
BufferedStreams [e1450e63] log:
 ├─possible versions are: 1.0.0 or uninstalled
 ├─restricted by compatibility requirements with MAT [23992714] to versions: 1.0.0
 │ └─MAT [23992714] log:
 │   ├─possible versions are: [0.5.0-0.5.1, 0.6.0, 0.7.0] or uninstalled
 │   └─restricted to versions * by an explicit requirement, leaving only versions [0.5.0-0.5.1, 0.6.0, 0.7.0]
 └─restricted by compatibility requirements with Compat [34da2185] to versions: uninstalled — no versions left
   └─Compat [34da2185] log:
     ├─possible versions are: [1.0.0-1.0.1, 1.1.0, 1.2.0, 1.3.0, 1.4.0, 1.5.0-1.5.1, 2.0.0, 2.1.0, 2.2.0, 3.0.0, 3.1.0, 3.2.0, 3.3.0-3.3.1, 3.4.0, 3.5.0, 3.6.0] or uninstalled
     └─restricted to versions 3.6.0-3 by DistributionsAD [ced4e74d], leaving only versions 3.6.0
       └─DistributionsAD [ced4e74d] log:
         ├─possible versions are: 0.4.1 or uninstalled
         └─DistributionsAD [ced4e74d] is fixed to version 0.4.1

```
```
My packages are as follows:
  [4fba245c] ArrayInterface v2.4.1
  [c52e3926] Atom v0.12.6
  [6e4b80f9] BenchmarkTools v0.4.3
  [76274a88] Bijectors v0.5.3
  [336ed68f] CSV v0.5.24
  [49dc2e85] Calculus v0.5.1
  [da1fd8a2] CodeTracking v0.5.8
  [5b389179] CompressedSensing v0.0.0 #master (https://github.com/dahlend/CompressedSensing.jl.git)
  [3a865a2d] CuArrays v1.7.2
  [864edb3b] DataStructures v0.17.9
  [31c24e10] Distributions v0.22.4
  [ced4e74d] DistributionsAD v0.4.1
[...]

PR #137, opened 2020-03-15
bump HDF5 requirement to include 0.13

bjarthur:
might fix https://github.com/JuliaIO/MAT.jl/issues/136

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,8 @@ os:
     - linux
     - osx
 julia:
-    - 0.7
-    - 1.0
-    - 1.1
     - 1.3
+    - 1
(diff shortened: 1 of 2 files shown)


bjarthur:
oh gosh.  HDF5 0.13 requires julia 1.3.  do we want to bump the julia requirement for MAT as well?

bjarthur:
any objections to dropping support for julia 1.0 like HDF5 did?

jaakkor2:
My system upgraded HDF5.jl to v0.13.0 and at the same time downgraded MAT.jl to v0.6.0. I "vote" for this PR.

PR #138, opened 2020-04-13
Update [compat] to allow CodecZlib v0.7

jaakkor2:


timholy:
Thanks!

timholy:
I'll make a release, but note it will only support Julia 1.3 or higher. If you need this on earlier releases we could create a branch, although there hasn't been a lot of maintenance activity here and I'd hate to add to the burden.

jaakkor2:
Thanks.  CodecZlib v0.7 requires Julia v1.3 as well.

Issue #140, opened 2020-08-03
MethodError reading v7.3 file


adamkapor:
I just did a clean install of Julia 1.5.  I get the following error trying to read a v7.3 file:

```
julia> MCMC2011 = MAT.matread(datapath*"/export_to_julia_2011.mat")["MCMC2011"]
ERROR: MethodError: no method matching h5d_read(::Int64, ::Int64, ::Array{Bool,2}, ::HDF5Properties)
Closest candidates are:
  h5d_read(::Any, ::Any, ::Any, ::Any, ::Any, ::Any) at C:\Users\akapor\.julia\packages\HDF5\QXcoh\src\HDF5.jl:2302
  h5d_read(::Int64, ::Int64, ::AbstractArray) at C:\Users\akapor\.julia\packages\HDF5\QXcoh\src\HDF5.jl:2057
  h5d_read(::Int64, ::Int64, ::AbstractArray, ::Int64) at C:\Users\akapor\.julia\packages\HDF5\QXcoh\src\HDF5.jl:2057
Stacktrace:
 [1] read(::HDF5Dataset, ::Type{Array{Bool,N} where N}) at C:\Users\akapor\.julia\packages\MAT\2LFMT\src\MAT_HDF5.jl:625
 [2] m_read(::HDF5Dataset) at C:\Users\akapor\.julia\packages\MAT\2LFMT\src\MAT_HDF5.jl:182
 [3] m_read(::HDF5Group) at C:\Users\akapor\.julia\packages\MAT\2LFMT\src\MAT_HDF5.jl:242
 [4] read(::MAT.MAT_HDF5.MatlabHDF5File, ::String) at C:\Users\akapor\.julia\packages\MAT\2LFMT\src\MAT_HDF5.jl:261
 [5] read(::MAT.MAT_HDF5.MatlabHDF5File) at C:\Users\akapor\.julia\packages\HDF5\QXcoh\src\datafile.jl:45
 [6] matread(::String) at C:\Users\akapor\.julia\packages\MAT\2LFMT\src\MAT.jl:133
 [7] top-level scope at none:1
```

Pkg.status says I have the following versions of things:
```
julia> Pkg.status()
Status `C:\Users\akapor\.julia\environments\v1.5\Project.toml`
  [c52e3926] Atom v0.12.19
  [31c24e10] Distributions v0.23.8
  [f67ccb44] HDF5 v0.13.3
  [e5e0dc1b] Juno v0.8.3
[...]

PR #141, opened 2020-08-06
Pass id rather than object as the 5th argument to h5d_read

grero:
This should fix issue #140.

### DIFF
--- a/src/MAT_HDF5.jl
+++ b/src/MAT_HDF5.jl
@@ -622,7 +622,7 @@ function read(obj::HDF5Dataset, ::Type{Array{Bool}})
     end
     # Use the low-level HDF5 API to put the data directly into a Bool array
     tf = Array{Bool}(undef, size(obj))
-    HDF5.h5d_read(obj.id, HDF5.hdf5_type_id(UInt8), tf, obj.xfer.id)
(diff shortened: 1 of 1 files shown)


src/MAT_HDF5.jl
```diff
@@ -622,7 +622,7 @@ function read(obj::HDF5Dataset, ::Type{Array{Bool}})
     end
     # Use the low-level HDF5 API to put the data directly into a Bool array
     tf = Array{Bool}(undef, size(obj))
-    HDF5.h5d_read(obj.id, HDF5.hdf5_type_id(UInt8), tf, obj.xfer)
+    HDF5.h5d_read(obj.id, HDF5.hdf5_type_id(UInt8), tf, obj.xfer.id)
```
jmert, line 625:
```suggestion
    HDF5.h5d_read(obj.id, HDF5.hdf5_type_id(UInt8), tf, convert(HDF5.Hid, obj.xfer))
```

Directly accessing `obj.xfer.id` will require bounding HDF5 to v0.13.2 or greater. The convert route preserves compatibility with older versions of HDF5.

jmert:
CI failures are solved by #142.

musm:
Does this need to also go in  0.8.1 ?

jmert:
To avoid the deprecation warnings completely, yes, it should go in the next release. Thanks!

PR #142, opened 2020-09-01
Fix how properties are passed during dataset creation

jmert:
JuliaIO/HDF5.jl#631 fixed a method ambiguity that is recognized in Julia v1.3+, but the fix caused the usage here to break. This duplicates the property creation/passing method used in the next `m_writearray` method.

Note that there are other breaking changes on recent HDF5.jl, which will require a pending PR (JuliaIO/HDF5.jl#655) to add in backwards-compatibility deprecations.


jmert:
I think this is also the fix for the problem mentioned in https://github.com/JuliaIO/MAT.jl/pull/141#issuecomment-678268989.

jmert:
A new HDF5.jl v0.13.6 has been tagged which introduces a deprecation for the method that #141 fixes, and this PR is the remaining fix necessary due to an ambiguity fix in HDF5's `d_create` methods.

PR #145, opened 2020-09-18
Adapt to changes in HDF5 v0.14 release

jmert:
[HDF5.jl v0.14.0](https://github.com/JuliaIO/HDF5.jl/releases/tag/v0.14.0)  has made a number of significant changes to the interface — this PR updates MAT for those changes.


musm (approved):

jmert:
Gah, I've seen the Linux x64/Julia 1.5.3 CI failure (https://github.com/JuliaIO/MAT.jl/runs/1449246460) a couple of times while rebasing this PR, but I cannot reproduce locally and it went away when I temporarily stuck `@show HDF5.checkvalid(x)` statements on `parent`, `dtype`, and `stype` [here](https://github.com/JuliaIO/MAT.jl/blob/e964bf068ce9937a80b6ecf6e1ae94cd094cdadb/src/MAT_HDF5.jl#L359).

musm:
Hmm, interesting. Should I hold off on the registration you think?

jmert:
I was hoping not, but it probably deserves a bit of investigation. I'll try running the tests a few more times repeatedly on my Linux machines to see if I can trigger, but I'm afraid it's going to be some subtle thing like somehow a finalizer is running early and could be hard to reliably reproduce. If you have a Linux machine of your own (especially if Ubuntu like CI; mine are Arch), maybe you could try running the tests a few times?

musm:
I won't cancel the registration PR since this looks intermittent and could be addressed in a patch release.

I'll try to see if I can trigger it, although it would be using Ubuntu through WSL2.

jmert:
Forcefully running the GC allows me to reproduce the error:
```diff
diff --git a/src/MAT_HDF5.jl b/src/MAT_HDF5.jl
index 87be6be..e6cf144 100644
--- a/src/MAT_HDF5.jl
+++ b/src/MAT_HDF5.jl
@@ -356,6 +356,10 @@ function m_writearray(parent::HDF5Parent, name::String, adata::AbstractArray{Com
     try
         stype = dataspace(adata)
         if compress
+            GC.gc(true)
+            @show parent
+            @show dtype
+            @show stype
             obj_id = create_dataset(parent, name, dtype, stype;
                               compress = 3, chunk = HDF5.heuristic_chunk(adata))
         else
```
`dtype` is closed/invalid after the GC sweep.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.3'
+          - '1'
```
musm, line 14: I don't think this actually runs the latest release version (i.e 1.5)  (unlike travis)

musm: I guess it now does. Maybe it was fixed, but I was sure it did not several months ago when transitioning HDF5.jl

jmert: From the CI logs, it looks like it does — it reports downloading Julia v1.5.3: https://github.com/JuliaIO/MAT.jl/pull/145/checks?check_run_id=1449246460#step:3:6

PR #151, opened 2020-11-24
Fix incorrect wrapping of a Datatype

jmert:
This was noticed in the CI of #145.

`create_datatype` already returns an HDF5.Datatype object, so by
wrapping again, the finalizer on the original object was allowed to
run too early.


musm:
Good catch!

PR #164, opened 2021-08-06
WIP: read (some) v4 format files

kleinschmidt:
This is based on the spec I found here: https://www.eiscat.se/wp-content/uploads/2016/03/Version-4-MAT-File-Format.pdf  It doesn't handle everything but maybe it's better than nothing!

Unfortunately I lack a sufficiently diverse collection of .mat files to test this against but it works for the one type of file I do have a lot of (a single `Int16` matrix per .mat file).

### DIFF
--- a/src/MAT.jl
+++ b/src/MAT.jl
@@ -28,6 +28,7 @@ using HDF5, SparseArrays
 
 include("MAT_HDF5.jl")
 include("MAT_v5.jl")
+include("MAT_v4.jl")
 
 using .MAT_HDF5, .MAT_v5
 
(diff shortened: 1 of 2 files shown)


ViralBShah:
@kleinschmidt Apologies for a ping after such a long time, but is it ok to merge this as is?

kleinschmidt:
Don't see why not, but to be perfectly honest I haven't touched a MAT file in many years

kleinschmidt:
Ah hold up, there's #186

matthijscox:
This PR does not seem to add any new features, so let's close it.

PR #181, opened 2023-02-28
Read Modelica MATv4 files

bc0n:
This adds support for reading v4 MAT files written by OpenModelica, according to the format detailed [here](https://openmodelica.org/doc/OpenModelicaUsersGuide/latest/technical_details.html#the-matv4-result-file-format).  This Modelica format is not readable by either #132 or #164.

At this point I'm not exporting any of the operations into MAT.jl.  As in test/runtests_modelica.jl they can be accessed directly by `MAT.MAT_v4_Modelica.readAclass("matfile.mat")`.  The function design is very literal right now and will be condensed into all-in-one accessors at some point.

Given the size of MAT files, read operations are atomic.  This differs from the MAT.jl interface, as the file pointer is never given to the user.  And right now, there is now ability to write files, mostly because I have no workflow requirement for this.  Also my development only included files authored by OpenModelica; Dymola appears to write a slightly different [format](https://www.claytex.com/tech-blog/trajectory-file-what-is-it-dissecting-a-dymola-result-file/).

Stepping back, I'd appreciate comments on the design and whether and how the operations for reading OpenModelica MATv4 files can be added to MAT.jl.  I was not able to use MAT.jl's basic accessors to read these files because of the [explicit v4 test](https://github.com/JuliaIO/MAT.jl/blob/f7302240f530b105db294c66f7740c92a7c79bbd/src/MAT.jl#L58).


sjkelly:
We now have the `version` kwarg to the `read/write` functions. I am not sure of the Matlab v4 and OpenModelica v4 distinction, so if there needs to be a slight different handling, we can add it as a special case.

sjkelly:
If we can autovdetect the Matlabv4 and the Modelicav4, I think that would be great. I think this PR certainly needs two things in the interim:

- Rebase on the latest merge for v4 support
- Drop Dataframes (This seems like a heavy dep for this package)

bc0n:
Thanks @qnikil7 et all, can you also simulate the [FallingBodyBox](https://github.com/mechanomy/MAT.jl/tree/master/test/v4_Modelica/FallingBodyBox) model under Dymola2021 so that we have some coverage into the Modelica Standard Library?

bc0n:
@sjkelly @qnikil7 I recall trying to use the deprecated v4 reader on Modelica files, but something about how it read the file and the assumed data type persuaded me to roll my own.  Kwargs are easy but there may be a better MAT.jl-level design decision to make that more flexibly reads differently-formatted files, something involving schemas?

qnikil7:
@bc0n

Created a [PR](https://github.com/mechanomy/MAT.jl/pull/2): added [FallingBodyBox](https://github.com/mechanomy/MAT.jl/tree/master/test/v4_Modelica/FallingBodyBox) model under Dymola2021, Removed Pkg dependency.

Also couple of thoughts
- I think `readVariable` and `readVairables`, can return a dictionary instead of a DataFrame, both of these functions being useful and easier to call.
 - For example While using the function `MAT.MAT_v4_Modelica.readVariable(ac, vn, vd, di, "time")` to get data for each variable, needs 4 other inputs. Would probably be helpful to have a function which which has two inputs, mat-file-path and variable name. (`readVariables` did that)

bc0n:
This is ready @sjkelly, the CI fails are in HDF5.jl and _memcpy @latest.

And from before, how or whether this integrates with the open-file approach of the other functions is open to discussion.  I added `isMatV4Modelica()` to support an integration into `read()`, but again the file sizes encourage retrieving specific variables only when needed.

sjkelly:
Thanks! Can you rebase this on master? The diff is difficult to read. If you need help rebasing, I am happy to do it for you.

bc0n:
@sjkelly the rebase was a bit rough but is much clearer now, let me know if you'd like me to open a new pr in place of this one.

src/MAT.jl
```diff
@@ -53,12 +54,6 @@ function matopen(filename::AbstractString, rd::Bool, wr::Bool, cr::Bool, tr::Boo
         return MAT_v4.matopen(rawfid, swap_bytes)
     end
 
-    # Test whether this is a MAT file
```
musm, line 56: reason to remove this?

.gitignore
```diff
@@ -1,5 +1,6 @@
 *~
 Manifest.toml
+desktop.ini
```
musm, line 3: is this needed?

bc0n: This ignores Windows' search breadcrumb files, not a big deal.

bc0n:
Looks like the _mecpy error has gone away; again let me know if there's anything I can do to get this to merge.  Or if it's not a fit I'll put it elsewhere.
Thanks

ViralBShah:
Given the size of this commit, wouldn't it be better to have a separate package?

bc0n:
That's fine, @ViralBShah, moving to https://github.com/mechanomy/ModelicaMAT

ViralBShah:
@bc0n Thanks! If you'd like to have that package in this org (naturally with ownership permissions), please let me know.

PR #198, opened 2024-02-02
Attempt reading old-style Matlab class in v7.3 as struct

Jeroen-van-der-Meer:
If you save an old-style Matlab object (i.e. created in an `@ClassName` folder, before `classdef` was a thing) in a `v7.3` file, `MAT.jl` throws an error saying it cannot read from a non-struct group (cf. line 232 of `MAT_HDF5.jl`). As a matter of fact, it appears that Matlab saves these objects 'as if' they are structs, and indeed the objects that I have appear to be read just fine if I manually remove the error and let the code continue.

Instead of throwing an error, I would propose giving a warning, such as

```julia
@warn "Unknown non-struct group of type $mattype detected; attempting to read as struct"
```

in case it can, in fact, be read properly.


ViralBShah:
@Jeroen-van-der-Meer Would it be useful for you to have commit access to this package, so that merges don't have to wait this long? I just happened to notice this PR and it seemed straightforward to merge.

musm:
I think we just forgot about it

PR #202, opened 2025-03-07
mat v4 compat: read namelen with given length and strip trailing null

songjhaha:
we have some old mat file，but the name is store without null-terminator '\0'.

MATLAB and scipy is compat with these file.

test .m script is:

```matlab
fid = fopen('testnamelen_4.2c_SOL2.mat', 'wb')
data = rand(1,9);
[M, N] = size(data);
data_type = 0;
prec = 'double';
name = 'testnamelen';

L = length(name);

fwrite(fid, data_type, 'int32');

fwrite(fid, M, 'int32');

fwrite(fid, N, 'int32');
fwrite(fid, 0, 'int32');
fwrite(fid, L, 'int32');
fwrite(fid, name, 'uint8');
fwrite(fid, data, prec);

data1 = load('testnamelen_4.2c_SOL2.mat');

assert(all(data1.(name)==data))
```


songjhaha:
@ViralBShah could you review this PR?

ViralBShah:
I do not know this package well enough to review - but it would be great for some of the new PR authors to come together to maintain the package.

ViralBShah:
Actually this looks straightforward enough. Should a test be added to actually test the added file?

songjhaha:
Tests have been added.

ViralBShah:
Happy to give you commit access if you can tag and release a new version etc.

Issue #201, opened 2025-02-13
TimeStamp support


aTrotier:
Hi, I have files from a motion databases which associates a motion matrice to a TimeStamp.

The timestamp parts is not loaded by MAT.jl
![Image](https://github.com/user-attachments/assets/eeed5bdc-0f79-4c69-9991-fbfb2b47b0fe)

The matlab type is "duration"

stemann:
Cf. `CompoundPeriod` for a possible type to translate into: https://docs.julialang.org/en/v1/stdlib/Dates/#Dates-and-Time-Types

aTrotier:
Right now, MAT.jl returns an empty value. I have to dig into the package to see where to put the conversion

aTrotier:
@stemann so I tried to load the data as UInt32 an that gives me : 
```
julia> vars["dur_1"]
6×1 Matrix{UInt32}:
 0xdd000000
 0x00000002
 0x00000001
 0x00000001
 0x00000001
 0x00000001

julia> Int.(ans)
6×1 Matrix{Int64}:
 3707764736
          2
          1
          1
          1
          1
```

No idea what to do with that to get that : 

![Image](https://github.com/user-attachments/assets/ea86e1f6-a2de-4b4a-870f-6da1f081f094)

Weirdly If I extract only one element of the TimeStamp I still get the same results...

stemann:
You may want to look into how `duration` is represented in the MAT-file - it looks like it could be a bit complicated: https://www.mathworks.com/help/coder/ug/define-duration-array-inputs.html

aTrotier:
I did it but no information about it.

I guess I have to reverse engineer it if I really want that possibility...

matthijscox:
We're working on class support, hopefully that will allow duration support soon as well: #205

matthijscox:
Reading of duration is supported now with the `Dates.Millisecond` type in v0.11

Issue #194, opened 2023-10-24
cannot read tables that are nested into a struct


lmulder1991:
I tried to load an object that I converted to a struct. The struct is correctly parsed into a Dict, but the table itself gets converted into another Dict with the "S0/1/2" & "arr" keys instead of the table it self. In the exampe I created a simple table, but it happens to any table I think, at least also my more complex table ends up in this same format.  

```julia
julia> using MAT

julia> file = matread(raw"path_to\test_data.mat")
Dict{String, Any} with 2 entries:
  ""  => UInt8[0x00 0x01 … 0x00 0x00]
  "s" => Dict{String, Any}("data"=>Dict{String, Any}("s1"=>Int8[77, 67, 79, 83], "arr"=>("", UInt32[0xdd000000; 0x00000002; … ; 0x00000001; 0x00000002;;]), "s0"=>Int8[], "s2"=>Int8[116, 97, 98, 108, 101]))

julia> file["s"]["data"]
Dict{String, Any} with 4 entries:
  "s1"  => Int8[77, 67, 79, 83]
  "arr" => ("", UInt32[0xdd000000; 0x00000002; … ; 0x00000001; 0x00000002;;])
  "s0"  => Int8[]
  "s2"  => Int8[116, 97, 98, 108, 101]
```

[test_data.zip](https://github.com/JuliaIO/MAT.jl/files/13111939/test_data.zip)

mkitti:
To clarify, what kind of "table" would you like to see? A `DataFrame` from DataFrames.jl?

lmulder1991:
Yes; that would work for me

KeithWM:
Facing the same/similar issue here with tables nested within _tables_. As an example, consider the attached file subtable.zip (containing a subtable.mat).
[subtable.zip](https://github.com/JuliaIO/MAT.jl/files/14486161/subtable.zip)
It was created, I think, with the following MNWE script:

```matlab
x = [11; 12];
y = [13; 14];
t1 = table(x, y);
x = [21; 22];
y = [23; 24];
t2 = table(x, y);
s.subtable = t1;
s(2).subtable = t2;
```

It contains a table, of which one of the fields is itself a table. Upon reading it in Julia, the result is not what I would hope:
```matlab
>> loaded = load('subtable.mat');
>> loaded.s(1).subtable
 
ans =
 
  2×2 table
 
    x     y 
    __    __
 
    11    13
    12    14
 
but in Julia the result is weird...
 
[...]

Issue #166, opened 2021-10-23
MAT.jl can not open variables storing string arrays. 


Leon6j:
I have two string arrays stored in a MAT file named test3.mat. They can be read out easily within Matlab, but I’m having big trouble reading them out in Julia.

Below is my code:

```
using MAT
F1 = matopen("test3.mat");
a = read(F1, "a");
a = read(F1, "b");
close(F1);
```

Here is a link to the test3.mat file:
`https://www.icloud.com/iclouddrive/0CtmgoektpD9oYZf2nzrN1aVQ#test3`

Here is one of the string arrays that was read out of the MAT file in Matlab:

a = 

  20×1 string array

    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"
    "HLY0802"

matthijscox:
MATLAB string class support is now available in MAT v0.11

Issue #158, opened 2021-03-03
Julia DateTime cannot be written in mat file


pablosaa:
I got the following error by trying to store Julia's DateTime variables into a MAT file, is there any work around for the error described below?

```
julia> using MAT
julia> using Dates

julia> t = now()
julia> fmat = matopen("test.mat", "w")
MAT.MAT_HDF5.MatlabHDF5File(HDF5 data file: test.mat, true, true, 0, false)

julia> write(fmat, "T", t)
ERROR: This is the write function for CompositeKind, but the input doesn't fit
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] m_write(::MAT.MAT_HDF5.MatlabHDF5File, ::HDF5.HDF5File, ::String, ::DateTime) at /home/user/.julia/packages/MAT/2LFMT/src/MAT_HDF5.jl:529
 [3] write(::MAT.MAT_HDF5.MatlabHDF5File, ::String, ::DateTime) at /home/user/.julia/packages/MAT/2LFMT/src/MAT_HDF5.jl:545
 [4] top-level scope at REPL[291]:1
```

PR #205, opened 2025-08-20
Update handling of mxOPAQUE_CLASS objects and Subsystem in MAT_v5.jl

foreverallama:
Spinoff from #23 to read the undocumented datatype `mxOPAQUE_CLASS`. Part of a series of changes to read and write these types across formats `v5` and `v7.3`. 

Some context :
- Subsystem data is written as a `uint8` array. However, this looks like another MAT-file that needs to be converted and read into. This needs to parsed before reading any variables in file.
- mxOPAQUE_CLASS variables are written with the following headers - Flags, Variable Name, Type Name, Class Name and Metadata
- MATLAB uses the same subsystem format for both v7 and v7.3 files, so starting with v7 is good enough.

## Edit:
Added a new file "MAT_subsys.jl" which contains methods for caching, parsing, and retrieving subsystem data to be assigned to an object. With this it should successfully load classdef objects. Additional context regarding how subsystem data is organized below:
- MCOS subsystem data is a cell array tagged to a class called "FileWrapper__"
- The first cell is a metadata array. It contains 9 blocks of metadata. Most of these blocks are to be interpreted as `uint32` integers even though its written in as `uint8`
-- Block 1 is a version indicator and some offset values
-- Block 2 is a list of class and property names as uint8 integers (null terminated)
-- Block 3 is a list of class IDs
-- Blocks 4 and 6 contain some metadata about how linking property names and property values
-- Block 5 is a list of object ID metadata
-- Block 7 is a list of dynamic properties attached to the object
-- Blocks 8 and 9 are unknown
- Cell 2 is empty (probably reserved?)
- Cell 3:end-3 are property values (depending on subsystem version it could be up to end-2)
- The last 2 or 3 cells are some kind of shared class templates. Only the last cell is known - it contains default property values

### DIFF
--- a/src/MAT_subsys.jl
+++ b/src/MAT_subsys.jl
@@ -22,6 +22,9 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+# For reference
+# https://github.com/foreverallama/matio/blob/main/docs/subsystem_data_format.md
+
 module MAT_subsys
 
 import ..MAT_types: MatlabStructArray, MatlabOpaque, convert_opaque
@@ -29,6 +32,7 @@ import ..MAT_types: MatlabStructArray, MatlabOpaque, convert_opaque
 export Subsystem
 
 const FWRAP_VERSION = 4
+const MCOS_IDENTIFIER = 0xdd000000
 
 mutable struct Subsystem
     object_cache::Dict{UInt32,MatlabOpaque}
@@ -264,7 +268,7 @@ function update_nested_props!(prop_value::Array{UInt32}, subsys::Subsystem)
     # MATLAB probably uses some kind of placeholders to decode
     # But this should work here
 
-    if first(prop_value) == 0xdd000000
+    if first(prop_value) == MCOS_IDENTIFIER
         # MATLAB identifies any uint32 array with first value 0xdd000000 as an MCOS object
         return load_mcos_object(prop_value, "MCOS", subsys)
     else
@@ -288,7 +292,8 @@ function get_saved_properties(
         elseif prop_type == 2
             prop_value = prop_field_idxs[i * 3 + 3]
         else
-            error("Unknown property type ID: $prop_type encountered during deserialization")
+            @warn "Unknown property type ID: $prop_type for property $prop_name encountered during deserialization"
+            prop_value = prop_field_idxs[i * 3 + 3]
         end
         save_prop_map[prop_name] = update_nested_props!(prop_value, subsys)
     end
@@ -360,7 +365,7 @@ function load_mcos_object(metadata::Array{UInt32}, type_name::String, subsys::Su
         return metadata
     end
 
-    if metadata[1, 1] != 0xDD000000
+    if metadata[1, 1] != MCOS_IDENTIFIER
         @warn "MCOS object metadata is corrupted. Returning raw data."
         return metadata
     end


foreverallama:
With these changes, full support is added for loading classdef objects in MAT-files in both v5 and HDF5 formats. Classdef objects are returned as a `Matrix{Dict{String, Any}}`. The `Dict` is a property `name, value` dictionary, with an additional key `__class__` containing the class name as a `String`.

The changes support different MAT-file and subsystem versions. It also supports loading all types of `MCOS` classdef objects (which is most of them), including handle class objects. Some other types I've seen are `java` and `handle` (for COM objects) which I don't know how to decode yet, but these are quite rare anyways and probably extremely specific to MATLAB.

Some notes:
1. I didn't really add a separate test because `test/v7.3/struct_table_datetime.mat` already seems to contain several objects like `datetime` `string` `categorical` and `table`. Just updated the test there instead

2. I just copied the copyright notice template from a different file, but I'm not too sure about adding a copyright notice for MAT_subsys.jl since the code is derived from reverse engineering the file format. Maybe someone else can comment on this?

3. For most of the classes like `datetime` or `string`, you will still need to decode the property map into usable information. It would be good to have some utility functions to do that. I've already [documented](https://github.com/foreverallama/matio/blob/v0.4.1/docs/field_contents.md) most of it, I'll get to it some other time though (or maybe someone else may take it up)

Edit: 
Squashed and consolidated changes for readability. Adds support for loading mxOPAQUE_CLASS objects from both v7 and v7.3 formats. For a review, the main part of the code would be `MAT_subsys.load_subsys!` and `MAT_subsys.load_mcos_object`. I've also highlighted some parts I'm not sure about with `FIXME` or `TODO`

matthijscox:
Hi there, I like your PR and I hope you still want to work on it. 

I have my own [PR-207](https://github.com/JuliaIO/MAT.jl/pull/207) and I will use `Array{Dict{String,Any}}` to write struct arrays, and maybe in the future we can also read struct arrays like that. I was wondering if there will be a conflict with your PR, since you will read (not write?) MATLAB classes as `Array{Dict{String,Any}}`. 

Is there a good way to separate normal struct arrays from MATLAB classes in the MAT.jl interface? Would the distinction come from the existence of the `"__class__"` key inside each `Dict` in the `Array{Dict{String,Any}}`?

(I also hope an active contributor/maintainer pops up for MAT.jl to review our work. I am trying to get in contact with them via email)

foreverallama:
Hey! 

It does look like there will be a conflict in our PRs. I aimed to provide read-write support in this module, but I only included read support for MATLAB classes in this PR to make it reasonable to review. 

Coming to the typing, my opinion is that the same types should be used for both read and write, and the typing should be consistent across the different MAT-file versions. I worked on a Pythonic version of this module, and there I used a wrapper class `MatioOpaque` to identify MATLAB classes instead. Something like that could be done here as well, or as you mentioned the `__class__` key (or some other special key) could be used as an identifier.

Ultimately, it's a design choice that would benefit from a discussion surrounding user requirements. It would be good if a maintainer (@ViralBShah ?) could pitch in their thoughts. As I understand there's not much activity here, but I would be happy to help out in any way I can!

matthijscox:
I like the idea of a wrapper type, at least for read/write consistency. Having obscure key names like `__class__` inside a Dict seems a little dodgy.

I wonder if it would be smart to have another wrapper type for the struct arrays? Some kind of mutable named tuple like type would be great, like: `Array{MatlabStruct{(:field1, :field2)}}`. And/or maybe actually use [StructArrays.jl](https://github.com/JuliaArrays/StructArrays.jl)

Looking at matio for inspiration, I found a [struct array test](https://github.com/foreverallama/matio/blob/main/tests/test_struct_cell.py#L382-L408), but I cannot quickly infer what Python type you are reading/writing? How do you differentiate between _cell arrays of structs_ and _struct arrays_? Right now MAT.jl does not have read/write consistency for struct arrays (they get re-written as a struct with cell arrays per field name).

ViralBShah:
@matthijscox  is a maintainer and we can add others too.

foreverallama:
> How do you differentiate between cell arrays of structs and struct arrays?

So my main point was to keep it consistent with `scipy` which uses `numpy` arrays to represent MATLAB types. Struct and Cells are both numpy arrays. The only difference is structs use numpy record arrays with each field having an `object` dtype, whereas Cell arrays have `dtype=object`. 

For example,
```
struct_arr_dtype = [('field1', object), ('field2', object)]
cell_arr_dtype = object

arr = some_arr
if arr.dtype.hasobject:
    if arr.dtype.names:
        # struct array
    else:
        # cell array
```

The only place this doesn't work is for empty structs which don't have field names. In this case, I've used a wrapper class `EmptyMatStruct` that contains the shape of the empty struct.

> Having obscure key names like __class__ inside a Dict seems a little dodgy

I do agree with this. In fact, I went through several iterations in my Python module before arriving at the current data representations. The main problem I had was ensuring separate types for simple read-write operations whilst keeping it user-friendly and scalable for the newer MATLAB classdef-based types. This lead to me using a bunch of wrapper classes that keep a balance between allowing numpy array operations and differentiability.

Something of that sort can be done here if it aligns with the scope of this package, and I would be happy to help. I'm not familiar with Julia typing and will need some time, but in the meantime you can [read](https://github.com/foreverallama/matio/blob/main/docs/field_contents.md#matlab-to-python-conversion) the type conversions I used in `matio`, relevant code [here](https://github.com/foreverallama/matio/blob/main/matio/utils/matclass.py). I'll also take a look at `MAT.jl` and see how we can unify the schema. Once a data representation is fixed it should be pretty straightforward from there.

matthijscox:
I've begun reading the code. The classes seem to be read mostly here as dicts in [MAT_subsys.jl R285-R288](https://github.com/JuliaIO/MAT.jl/pull/205/files#diff-8085f027bcfb8e0d22c7f8ba11ece7a6348154b5972a3fd726e417c74ea48e8bR285-R288).

I suppose we could define a `MatlabOpaque` (or `MatlabClass` or `MatlabObject`) type that mirrors a dict:
```julia
struct MatlabOpaque{T}
    classname::Symbol
    properties::Vector{String} # or Vector{Symbol}
    values::Vector{T}
end
```

Optionally I wonder if we want to be able to expose the class "type" to the Julia type system. (These Val types do come with a small performance penalty I remember)

```julia
struct MatlabOpaque{C<:Val, T}
    properties::Vector{String} # or Vector{Symbol}
    values::Vector{T}
end
# outer constructor
function MatlabOpaque(classname::Symbol, properties::Vector{String}, values::Vector{T}) where T
   return MatlabOpaque{Val{classname}, T}(properties, values)
end
```

Now you can identify a vector of same class (class/struct array?) from a vector different classes (cell array?)
```julia
MatlabOpaque{Val{:table}, Any}[] # class array of Matlab tables
MatlabOpaque[] # class (cell) array of different classes
Any[] # cell array of any kind of (matlab) type
```

This also means we can write automatic converters:
```julia
Base.convert(::Type{DataFrame}, MatlabOpaque{Val{:table}}) = ...
```
or better yet add the [Tables.jl interface](https://tables.juliadata.org/stable/implementing-the-interface/):
```julia
Tables.istable(::Type{MatlabOpaque{Val{:table}}) = true
Tables.istable(::Type{MatlabOpaque}) = false
```

Ofcourse instead of the Val type parameter, we could just convert immediately to different known types during the reading. Then we have to define the multiple MAT.jl internal types we need, or immediately convert to the most sensible Julia type.
```julia
struct MatlabTable 
...
end
Tables.istable(::Type{MatlabTable}) = true

struct MatlabDuration <: Dates.AbstractTime # or use Dates.Millisecond type
...
end
```

I would propose to first implement the `MatlabOpaque` type (with or without Val parameter) in this PR and then in future PRs we can think about the (automatic) conversion to Julia types for `table`, `datetime`, `duration`, etc.

foreverallama:
If I'm understanding correctly, we'd have `MatlabOpaque{Val{:classname}, Any}[]` for MATLAB object arrays. So this resolves the conflict with struct arrays you proposed?

Some points to clarify/get clarified:
- The `MatlabOpaque` type would have two vectors for `properties` and `values`. Do we retain dictionary behaviour instead of indexing? Idk about the performance point of view, but `dict` behaviour would be helpful, especially for conversions to Julia types as MATLAB types can contain like 10-20 properties for complex types like `timetable`.
- Can we currently load/represent `1x0` `0x1` or `0x0` structs with the `Dict` type? We need that functionality to write classdef objects to MAT-files.
- I guess object arrays would then be `Matrix{MatlabOpaque{T}}(undef, dims)` right?
- What do you suggest for `mxOBJECT_CLASS` and `mx_FUNCTION_CLASS`. The serialized data for these types are saved as structs, and `MAT.jl` currently loads them as `Dict` with an extra key `class`.

The rest sounds good. Doesn't sound like it diverges too much from the current state of the PR. To summarize the main edits:
- Thread safe cache handling
- Change return type to `MatlabOpaque`

matthijscox:
> If I'm understanding correctly, we'd have MatlabOpaque{Val{:classname}, Any}[] for MATLAB object arrays. So this resolves the conflict with struct arrays you proposed?

Indeed.

Note I now created a `MatlabStructArray` type in my PR to fully avoid any conflict with your types I hope.

> The MatlabOpaque type would have two vectors for properties and values. Do we retain dictionary behaviour instead of indexing? Idk about the performance point of view, but dict behaviour would be helpful, especially for conversions to Julia types as MATLAB types can contain like 10-20 properties for complex types like timetable.

You mean that you want to index by name `object["name"]` or `object[:name]`? Coincidentally I added Dict-like behavior to my `MatlabStructArray` type (string-based indexing and iteration returns key-value pairs). You can have a look for inspiration: [see here for example](https://github.com/JuliaIO/MAT.jl/pull/207/files#diff-3ba5a23cd064cadaa0cff3990a64563928407a63c8e8096d1ac885107a9f7c12R70-R73).

Basically we can define methods like:
```julia
function Base.getindex(object::MatlabOpaque, prop::AbstractString)
   idx = findfirst(isequal(prop), object.properties)
   return object.values[idx]
end
```

> Can we currently load/represent 1x0 0x1 or 0x0 structs with the Dict type? We need that functionality to write classdef objects to MAT-files.

Pff, I will have to check. You also make me worried about these empty sized struct arrays for my PR.

Is this the way to construct them?
```
>> f = {'a', 'b', 'c'};
>> f{2,1} = {};
>> s = struct(f{:})

ans = 

  0×0 empty struct array with fields:

    a
    b
    c

>> reshape(s, 0, 1);
>> reshape(s, 1, 0);
```

> I guess object arrays would then be Matrix{MatlabOpaque{T}}(undef, dims) right?

Yeah, well `Matrix{MatlabOpaque{C,T}}(undef, dims)` if we choose the Val-based dispatching option. Though I'm thinking now that we can probably drop the `T` parameter entirely in the struct definition if it's always of type `Any`. Which seems likely since we don't often expect the same types between MATLAB properties?

> What do you suggest for mxOBJECT_CLASS and mx_FUNCTION_CLASS. The serialized data for these types are saved as structs, and MAT.jl currently loads them as Dict with an extra key class.

I very new to MAT.jl myself I admit. I saw the `mxOBJECT_CLASS` in the MAT_v5.jl file and I'm not sure what it is? Is it the old matlab object class created in `@classname` folders? I don't see it mentioned in MAT_HDF5.jl (v7.3).

----

It's possible we're going to have merge conflicts with my PR, since I've been refactoring a bit there. So if you have any time to review my PR that would be great.

matthijscox:
> Can we currently load/represent 1x0 0x1 or 0x0 structs with the Dict type? We need that functionality to write classdef objects to MAT-files.

To quickly follow-up on your question:

## The bad news

Currently (MAT v0.10) we cannot. An empty `0x0` struct array will be read as `Dict{String, Any}("c"=>Matrix{Any}(undef, 0, 0), "b"=>Matrix{Any}(undef, 0, 0), "a"=>Matrix{Any}(undef, 0, 0))` and written back as a `1x1` struct with empty cell arrays in the field:
```matlab
>> s00

s00 = 

  struct with fields:

    c: {}
    b: {}
    a: {}
```

## Good news
In my new struct array PR I can write them! (reading goes wrong, I'll fix that):

In Julia:
```julia
empty_sarr = MAT.MatlabStructArray(["a", "b", "c"], [Matrix{Any}(undef, 0, 0), Matrix{Any}(undef, 0, 0), Matrix{Any}(undef, 0, 0)])
matwrite("test.mat", Dict("s00" => empty_sarr))
```
In MATLAB:
```matlab
>> load('test.mat')
>> s00

s00 = 

  0×0 empty struct array with fields:

    a
    b
    c
```

foreverallama:
> You also make me worried about these empty sized struct arrays for my PR

Haha yeah, sorry. It's an edge case from a user perspective, but MATLAB uses these zero-element dimensions as placeholders, and getting them wrong results in unexpected errors when loading in MATLAB. Ideally if the `read_struct` and `write_struct` methods can take care of this I can simply use `MatlabStructArray` in subsystem. In fact it might be better to finish your PR first and then come back to this.

> we can probably drop the T parameter entirely in the struct definition

yeah, sounds good. 

> I saw the mxOBJECT_CLASS in the MAT_v5.jl file and I'm not sure what it is? Is it the old matlab object class created in @classname folders? I don't see it mentioned in MAT_HDF5.jl (v7.3).

I believe `mxOBJECT_CLASS` is legacy classdef implementation in MATLAB. I guess we can keep it as is, maybe add `MatlabFunction` and `MatlabObject` types later if required.

I did notice some differences between the HDF5 and v5 versions. The HDF5 file isn't as extensive as the v5 file, and I honestly think it needs a rewrite to allow consistency between versions. But that's for another time.

matthijscox:
I now support empty struct arrays in my PR #207. Create one easily via `MatlabStructArray(["a", "b"], (0,0))`

I also created the `MatlabClassObject` to better support the `mxClassObject` in v5 format and read-write the class object in v7.3 HDF5 format.
I allow class object arrays via `MatlabStructArray` if a class name is provided.

I think that's all we need to continue with your PR. I will probably merge soon and then if you want I can help with this PR to solve any merge conflicts for you.

foreverallama:
Thanks! I'm a bit swamped till the end of the month and will most probably take a look at this after, will try to squeeze something in as soon as I can

matthijscox:
Somehow I increased SnoopCompile invalidations a bit, this might mean package loading is slightly slower, but it's all due to dependencies (mostly HDF5.jl). Not sure I know how to fix this, so maybe we'll just take the hit.

On master branch, according to CI:
```
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬───────────────┬───────────────┬─────────────────┐
│ <file name>:<line number>                                                                                                            │ Function Name │ Invalidations │ Invalidations % │
│                                                                                                                                      │               │               │     (xᵢ/∑x)     │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────┼───────────────┼─────────────────┤
│ HDF5/Z859u/src/dataspaces.jl:204                                                                                                     │    convert    │      830      │      62.44      │
│ /cache/build/tester-amdci4-14/julialang/julia-release-1-dot-12/usr/share/julia/stdlib/v1.12/SparseArrays/src/readonly.jl:23          │   eachindex   │      146      │     10.984      │
│ HDF5/Z859u/src/datasets.jl:332                                                                                                       │     write     │      146      │     10.984      │
│ /cache/build/tester-amdci4-14/julialang/julia-release-1-dot-12/usr/share/julia/stdlib/v1.12/SparseArrays/src/solvers/spqr.jl:143     │     axes      │      51       │      3.838      │
│ /cache/build/tester-amdci4-14/julialang/julia-release-1-dot-12/usr/share/julia/stdlib/v1.12/SparseArrays/src/sparsevector.jl:1667    │  _mapreduce   │      36       │      2.709      │
│ /cache/build/tester-amdci4-14/julialang/julia-release-1-dot-12/usr/share/julia/stdlib/v1.12/SparseArrays/src/solvers/cholmod.jl:1354 │     size      │      30       │      2.258      │
│ HDF5/Z859u/src/filters/filters.jl:428                                                                                                │    convert    │      28       │      2.107      │
│ HDF5/Z859u/src/datasets.jl:703                                                                                                       │  IndexStyle   │      25       │      1.881      │
│ BufferedStreams/YvcPf/src/emptystream.jl:81                                                                                          │     write     │      20       │      1.505      │
│ BufferedStreams/YvcPf/src/bufferedinputstream.jl:152                                                                                 │     peek      │       6       │     0.4514      │
│ TranscodingStreams/O3BYF/src/stream.jl:245                                                                                           │     reset     │       3       │     0.2257      │
│ TranscodingStreams/O3BYF/src/stream.jl:229                                                                                           │   ismarked    │       2       │     0.1505      │
│ TranscodingStreams/O3BYF/src/stream.jl:234                                                                                           │     mark      │       2       │     0.1505      │
│ /cache/build/tester-amdci4-14/julialang/julia-release-1-dot-12/usr/share/julia/stdlib/v1.12/LibGit2/src/types.jl:1359                │     get!      │       2       │     0.1505      │
│ /cache/build/tester-amdci4-14/julialang/julia-release-1-dot-12/usr/share/julia/stdlib/v1.12/SparseArrays/src/sparsevector.jl:2292    │     sort!     │       1       │     0.07526     │
│ /cache/build/tester-amdci4-14/julialang/julia-release-1-dot-12/usr/share/julia/stdlib/v1.12/LibGit2/src/walker.jl:77                 │     sort!     │       1       │     0.07526     │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴───────────────┴───────────────┴─────────────────┘
```

On my current local commit:
```
┌──────────────────────────────────────────────────────────────────────────────────────────────┬───────────────┬───────────────┬─────────────────┐
│ <file name>:<line number>                                                                    │ Function Name │ Invalidations │ Invalidations % │
│                                                                                              │               │               │     (xᵢ/∑x)     │
├──────────────────────────────────────────────────────────────────────────────────────────────┼───────────────┼───────────────┼─────────────────┤
│ C:\\Users\\matcox\\.julia\\packages\\HDF5\\Z859u\\src\\dataspaces.jl:204                     │    convert    │      581      │      28.84      │
│ C:\\workdir\\usr\\share\\julia\\stdlib\\v1.12\\SparseArrays\\src\\readonly.jl:23             │   eachindex   │      418      │      20.75      │
│ C:\\Users\\matcox\\.julia\\packages\\StringEncodings\\unzPp\\src\\encodings.jl:17            │    convert    │      357      │      17.72      │
│ C:\\Users\\matcox\\.julia\\packages\\PooledArrays\\Vy2X0\\src\\PooledArrays.jl:499           │    convert    │      286      │      14.2       │
│ C:\\Users\\matcox\\.julia\\packages\\HDF5\\Z859u\\src\\attributes.jl:31                      │     write     │      92       │      4.566      │
│ C:\\Users\\matcox\\.julia\\packages\\HDF5\\Z859u\\src\\datasets.jl:703                       │  IndexStyle   │      75       │      3.725      │
│ C:\\workdir\\usr\\share\\julia\\stdlib\\v1.12\\SparseArrays\\src\\solvers\\spqr.jl:143       │     axes      │      51       │      2.533      │
│ C:\\workdir\\usr\\share\\julia\\stdlib\\v1.12\\SparseArrays\\src\\sparsevector.jl:1667       │  _mapreduce   │      38       │      1.887      │
│ C:\\workdir\\usr\\share\\julia\\stdlib\\v1.12\\SparseArrays\\src\\solvers\\cholmod.jl:1354   │     size      │      30       │      1.489      │
│ C:\\Users\\matcox\\.julia\\packages\\HDF5\\Z859u\\src\\filters\\filters.jl:428               │    convert    │      28       │      1.391      │
│ C:\\Users\\matcox\\.julia\\packages\\PooledArrays\\Vy2X0\\src\\PooledArrays.jl:513           │    convert    │      15       │     0.7446      │
│ C:\\workdir\\usr\\share\\julia\\stdlib\\v1.12\\SparseArrays\\src\\higherorderfns.jl:1188     │     map!      │      14       │     0.6953      │
│ C:\\Users\\matcox\\.julia\\packages\\BufferedStreams\\YvcPf\\src\\emptystream.jl:81          │     write     │      12       │     0.5957      │
│ C:\\Users\\matcox\\.julia\\packages\\BufferedStreams\\YvcPf\\src\\bufferedinputstream.jl:152 │     peek      │       6       │     0.2979      │
│ C:\\Users\\matcox\\.julia\\packages\\TranscodingStreams\\O3BYF\\src\\stream.jl:229           │   ismarked    │       4       │     0.1986      │
│ C:\\Users\\matcox\\.julia\\packages\\TranscodingStreams\\O3BYF\\src\\stream.jl:234           │     mark      │       4       │     0.1986      │
│ C:\\Localdata\\Repositories\\MAT.jl\\src\\MAT.jl:244                                         │     names     │       2       │     0.0993      │
│ C:\\workdir\\usr\\share\\julia\\stdlib\\v1.12\\SparseArrays\\src\\sparsevector.jl:2292       │     sort!     │       1       │     0.04965     │
└──────────────────────────────────────────────────────────────────────────────────────────────┴───────────────┴───────────────┴─────────────────┘

```

Seems I managed to suppress a few HDF5 invalidations (probably by defining a few more explicit typed `m_write` functions), but I gained a bunch of SparseArrays invalidations (confused about that one) and some for the new dependencies (StringEncodings and PooledArrays).

matthijscox:
Slightly confused right now. Locally on Julia 1.12.0 I have 883 invalidations. CI (on 1.12.1) says 1474 invalidations

Switched to Julia 1.12.1 with a fresh environment. Got 1040 invalidations

foreverallama:
I added tests for handle class objects and nested properties. I also added support for loading dynamic properties (with tests). Made some tiny changes to accommodate - object arrays are now `Array{MatlabOpaque}` instead of `Array{Any}`.

I think we should clean up this PR and focus on supporting more classes/types in a different one. I will add write support for `MatlabOpaque` after this PR is merged.

matthijscox:
I agree. Do you see any more cleanup needed? Because I'd love to release MAT.jl v0.11 after merging this PR.

matthijscox:
> object arrays are now Array{MatlabOpaque}

I think this will fail if one the objects in the array gets automatically converted, e.g. to a DateTime. I will add such a test case tomorrow.

foreverallama:
> Do you see any more cleanup needed?

I'll just add some comments to try and improve readability of `MAT_subsys.jl`. I might be a bit biased though and could use another perspective for suggestions. I updated workflows in `matio` after initially drafting this PR, I'll just make sure everything's aligned well. Any other linting issues can be corrected now. Good time to update docs and readme as well.

The invalidations stuff is pretty new to me. If I understand correctly, the number of invalidations in the PR must be less than that in master? If its not critical, can we bypass it and tackle it in a separate issue?

> I think this will fail if one the objects in the array gets automatically converted, e.g. to a DateTime

Ah so this is a unique quirk of MATLAB, and shouldn't be an issue. For user-defined classes, object arrays contain a separate object instance in each element of the array (unless they are handle class objects which may be references). Also, all objects in an array must be of the same class. But for MATLAB datatypes like `datetime` and others, it is always a scalar even if the underlying data is an array. MATLAB overloads assignment operations in the class definition, ensuring that only one object instance is created and that its corresponding properties store the array behaviour instead.

matthijscox:
> The invalidations stuff is pretty new to me. If I understand correctly, the number of invalidations in the PR must be less than that in master? If its not critical, can we bypass it and tackle it in a separate issue?

Yeah I can merge with the "invalidations CI step" failing. Then I suppose we have a new baseline. 

> Ah so this is a unique quirk of MATLAB, and shouldn't be an issue. For user-defined classes, object arrays contain a separate object instance in each element of the array (unless they are handle class objects which may be references). Also, all objects in an array must be of the same class. But for MATLAB datatypes like datetime and others, it is always a scalar even if the underlying data is an array. MATLAB overloads assignment operations in the class definition, ensuring that only one object instance is created and that its corresponding properties store the array behaviour instead.

If this is true, that means we can actually skip the `convert_opaque` call in the for loop for `Array{MatlabOpaque}`

> Any other linting issues can be corrected now.

We could add a formatting style to MAT.jl, none has been chosen so far. I typically opt for [bluestyle](https://domluna.github.io/JuliaFormatter.jl/dev/blue_style/). We can at least reformat our new files in this PR.

> I'll just add some comments to try and improve readability of MAT_subsys.jl. I might be a bit biased though and could use another perspective for suggestions.

The MCOS system seems obscure and the code obvious reflects that a bit. I'm just happy it works at all. We can refactor and comment on the obscure parts, but I'm okay with what we have. (Though what the on earth are some of these fields like `_u6_metadata` even doing)

matthijscox:
I took the liberty to refactor a bit and format the style for MAT_types.jl and MAT_subsys.jl

foreverallama:
Sounds good. Everything with subsystem is reverse engineered so we can only guess as to what they do. Some fields are unknown still, but we still need them to write back to a MAT-file. Interestingly, MATLAB does not perform any sanity checks when loading subsystem data, so you can easily crash it by simply going out of bounds in array indexing. For safety, we could add sanity checks here if required. 

I wanted one thing clarified before we move forward. I noticed that both MATLAB char arrays and string arrays are returned as Julia strings. It would be better if we could somehow separate this behaviour. Char arrays are written as a bunch of ASCII or UTF-8 encodings, but string arrays use the MCOS opaque class format. It's not an issue in `read`, but it might be an issue when trying to differentiate during `write` operations. One of the issues I've come across earlier is that some properties in datatypes like `timetable` expect MATLAB strings, and will not work with char arrays in MATLAB. I resolved this issue in `matio` by using NumPy's `unicode` strings for char arrays (same as `SciPy`) and using the newer `StringDType()` for MATLAB strings. 

What are your thoughts on this?

Finally, I'm listing a bunch of issues we can close after merging this PR.
#201 #194 #166 #158 #133 #127 #43 
#82 can soon be closed as well
And finally we can close the other PR opened a long time back in #23 !

matthijscox:
> One of the issues I've come across earlier is that some properties in datatypes like timetable expect MATLAB strings, and will not work with char arrays in MATLAB. I resolved this issue in matio by using NumPy's unicode strings for char arrays (same as SciPy) and using the newer StringDType() for MATLAB strings.

I see. We can obviously postpone until the writing PR. We could use the same approach as you do, but then we need to search for another Julia string type other than the base UTF-8 `String`. Perhaps one of the [LegacyStrings](https://github.com/JuliaStrings/LegacyStrings.jl) types? And then we have to decide which of the two (char arrays or string classes) is used for the base `String`. I suppose char arrays are UTF-8?

matthijscox:
Seems to me this PR is ready. Let me know if you are happy with the code or want to do further refactoring. Once you give the go, I'll merge and register v0.11.

foreverallama:
Lets go for it. We can open new issues for the other stuff, they won't affect base functionality.

> I suppose char arrays are UTF-8?

I think they're a mix of ASCII or UTF-8, even though MATLAB documentation says they [use UTF-16](https://www.mathworks.com/help/matlab/matlab_prog/unicode-and-ascii-values.html). It might be optimized during save based on content? MATLAB strings though are definitely using UTF-16.

Issue #206, opened 2025-10-31
How to write an array of structs?


matthijscox:
I noticed that I can read normal arrays of structs and cell arrays of structs with MAT.jl, but I cannot write arrays of structs. At least not by default. Would love to know how to do that.

## MATLAB input data

Here's an example:
```matlab
s = struct();
s.array = [struct('x', [1,2], 'y', [3,4]), struct('x', [5,6], 'y', [7,8])];
s.cell = {struct('x', [1,2], 'y', [3,4]), struct('x', [5,6], 'y', [7,8])};
save('struct_test.mat', 's')
```

This is the input, note that we have a `1x2struct` array

```matlab
>> s

s = 

  struct with fields:

    array: [1×2 struct]
     cell: {[1×1 struct]  [1×1 struct]}

>> s.array

ans =

[1x2 struct, 608 bytes]
x: [1, 2]
y: [3, 4]
```

## Julia conversion

Let's read and write in Julia without editing:
```julia
using MAT
vars = matread("struct_test.mat")
matwrite("julia_struct_test.mat", vars)
```

Note that inside Julia, we already have something unexpected. I expected a struct/dict array for the MATLAB struct array, but instead we get arrays of arrays inside the `"x"` and `"y"` keys, e.g. see `"x" => Any[[1.0 2.0] [5.0 6.0]]`. 
In the cell array case we do have an array of dicts.

```julia
julia> vars["s"]
Dict{String, Any} with 2 entries:
  "array" => Dict{String, Any}("x"=>Any[[1.0 2.0] [5.0 6.0]], "y"=>Any[[3.0 4.0] [7.0 8.0]])
  "cell"  => Any[Dict{String, Any}("x"=>[1.0 2.0], "y"=>[3.0 4.0]) Dict{String, Any}("x"=>[5.0 6.0], "y"=>[…

julia> vars["s"]["array"]
Dict{String, Any} with 2 entries:
  "x" => Any[[1.0 2.0] [5.0 6.0]]
  "y" => Any[[3.0 4.0] [7.0 8.0]]

julia> vars["s"]["cell"]
1×2 Matrix{Any}:
 Dict{String, Any}("x"=>[1.0 2.0], "y"=>[3.0 4.0])  Dict{String, Any}("x"=>[5.0 6.0], "y"=>[7.0 8.0])
```

## MATLAB loading
[...]

PR #207, opened 2025-11-06
Struct array writing

matthijscox:
My attempt to solve https://github.com/JuliaIO/MAT.jl/issues/206 and allow struct array writing in HDF5 mat files.

Changes:
* defines `m_write` for `AbstractArray{<:AbstractDict}` as interface for struct arrays
* refactor cell array writing (`m_write` for `AbstractArray{T}`) to have common reference writing
* added readme example

### DIFF
--- a/src/MAT_types.jl
+++ b/src/MAT_types.jl
@@ -70,9 +70,8 @@ module MAT_types
         names::Vector{String}
         values::Vector{Array{Any,N}}
         class::String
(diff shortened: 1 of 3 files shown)


matthijscox:
Defined the `MatlabStructArray` type to enabled read-write consistency of struct arrays for v7.3 mat files. I'm trying to keep the type as backwards compatible as possible with the previous `Dict{String,Any}` type that contains arrays in the field, yet also allow easy conversion to arrays of dicts which is similar to cell array behavior.

I don't know yet how to distinguish this struct array in mat versions before v7.3.

PR #208, opened 2025-11-21
Writing MatlabOpaque objects to MAT-files

foreverallama:
Part of a series of commits to support write operations for `MatlabOpaque` objects. For now I just rearranged the methods to make it easy for me to update the code. 

I'll first add to `MAT_subsys.jl` before updating `MAT_HDF` and `MAT_v5`.

### DIFF
--- a/test/write.jl
+++ b/test/write.jl
@@ -216,6 +216,19 @@ end
     end
 end
 
+@testset "Empty Struct 1x1" begin
+    var_dict = Dict("empty_struct" => Dict{String,Any}())
+    mktempdir() do tmpdir
+        tmpfile = joinpath(tmpdir, "test.mat")
+        matwrite(tmpfile, var_dict)
+        read_var = matread(tmpfile)
+
+        @test haskey(read_var, "empty_struct")
+        @test isa(read_var["empty_struct"], Dict{String,Any})
+        @test length(keys(read_var["empty_struct"])) == 0
+    end
+end
+
 @testset "MatlabOpaque handle" begin
     d = Dict{String,Any}("a" => 1, "b" => Any[1.0, 2.0])
     obj = MatlabOpaque(d, "TestClassHandle")


foreverallama:
CC @matthijscox 

Added some methods to write subsystem data... WIP. Basically just translating Python code into Julia here, so would need a review.

I had to make some small adjustments to incorporate the write logic. I updated some metadata fields from `Vector{UInt32}` to `Vector{Vector{UInt32}}`. The main reason is that when serializing nested objects, full metadata for the outermost object is saved before saving metadata for any inner nested objects. The length of the metadata fields are variable, so instead of using a position-based update logic (where you might need to recalculate positions every time), I went with a mutable vector insertion instead. This results in the `Vector{Vector{UInt32}}` type. Since during load, we're only expecting one vector, I simply updated the loading logic to use `metadata_field[1][idx]` instead of `metadata_field[idx]`.

I don't think performance would take a hit when splatting these finally as we don't expect too many objects in a MAT-file (after splat the total FileWrapper metadata length could be around thousands at best for really complex class definitions). There might be a better way to code the whole logic but until then this should work. 

I also added a new object cache for saving that would key by `id(MatlabOpaque)`, and renamed the existing one to `load_object_cache`.

A side note, I was able to figure out what the `_c2` metadata field contains. It's something called a [class alias](https://www.mathworks.com/help/matlab/matlab_oop/class-aliasing.html). Accordingly I renamed the field `mcos_class_alias_metadata`. Essentially, this allows you to create an object instance using an alias. The `_c2` metadata field contains the latest alias name, which is also saved in `subsystem.mcos_names`. However, the object itself is saved by the oldest alias which would be stored in the `MatlabOpaque.class` attribute. I don't think its super important to support this feature here, but we could consider adding a new attribute `alias` to `MatlabOpaque` if required.

matthijscox:
Yeah this whole class system is beyond me :) 
I don't see any obvious issues with the code on first glance.

I do notice there are zero tests for the new code. I guess you cannot create a minimal write test yet?
I also don't know if we want to test any of the subsystem internals.

foreverallama:
Code's not complete yet,. Still working on it, trying to squeeze in a little when I'm free. No point in testing internals as the write tests would take care of it anyways

foreverallama:
Most of the code's ready. It's not working yet I'll debug soon and in the meantime add some sample tests

src/MAT_subsys.jl
```diff
@@ -394,4 +428,300 @@ function load_mcos_object(metadata::Array{UInt32}, type_name::String, subsys::Su
...
+
+function set_fwrap_data!(subsys::Subsystem)
+
+    fwrap_data = Any[]
+    fwrap_metadata = set_fwrap_metadata!(subsys)
+    push!(fwrap_data, reshape(fwrap_metadata, :, 1))
+    push!(fwrap_data, Any[])
+    append!(fwrap_data, subsys.prop_vals_saved)
+
+    empty_struct = MatlabStructArray([]) #! FIXME
```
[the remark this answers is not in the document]
matthijscox: Reading the code, this might not yet be possible. I always estimate the dimensions from the first column, but if there are no columns... Might be we have to explicitly add the dimensions to the MatlabStructArray type, so that it also works without any columns.

Does the empty struct array have any special features inside the HDF5 format?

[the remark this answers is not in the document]
foreverallama: > easy way to load an example fully empty (no columns, 1x0 matrices) struct array type

Sorry, yeah that's a good point. I'll check this out first

src/MAT_HDF5.jl
```diff
@@ -682,6 +689,20 @@ end
 
 # Write a struct from arrays of keys and values
 function m_write(mfile::MatlabHDF5File, parent::HDF5Parent, name::String, k::Vector{String}, v::Vector)
+    if length(k) == 0
```
foreverallama: I added the ability to write empty dicts as 0x0 structs over here. The subsystem data works with 0x0 structs as well in MATLAB. 

This eliminates the need for any new struct markers for now.

foreverallama:
It appears to be working now. I tried a minimal test below and it can be read in all of `MAT.jl`, `matio` and MATLAB.

```julia
sample_obj = MatlabOpaque(Dict(
    "a" => reshape([10.0], 1, 1),
    "b" => reshape([1.0, 2.0, 3.0], 3, 1)
), "TestClass")

sample_dict = Dict(
    "var" => sample_obj,
)

MAT.matwrite("test.mat", sample_dict)
```

I'll now focus on some simple tests covering user-defined objects and handle class objects. I still need to add support for object arrays. If I've missed anything else let me know

src/MAT_HDF5.jl
matthijscox: Alright. Are 0x0 structs also read as empty dicts?

foreverallama: Yup, all of 1x0, 0x1 and 0x0 are read as empty dicts already. But I hardcoded 0x0 for empty dict during write.

matthijscox: It's a rather special corner case. I'm okay with a new type that we keep internal for the writing, and then we can always consider to refactor later:
```julia
struct EmptyStruct
   dims::Vector{Int}
end
```

foreverallama:
* Added support for writing object arrays
* Added simple test for basic objects, handle class objects, object arrays and nested objects
* Empty Dict is now written in as a 1x1 struct with no fields (fixes #214 )
* New type `EmptyStruct` to write special zero-dimensional structs (for internal use)

The objects being written are loading successfully in MATLAB (2023b). With this I think the base for writing objects is covered. We can now write methods for converting from say `datetime` etc. to `MatlabOpaque`. A point to note is that subsystem data is being written in version 4. We can also write in v2 instead, though I think v4 has been the standard for a while now.  

There's a special case of zero-dimensional objects, but again its an edge case that can be tackled if required.

I also opened #213 to fix #212 , it would be easier to merge that first and then fix conflicts here.

matthijscox:
Do the tests now also cover the empty dict read/write case? If not, it would be good to add a test for that, so we don't accidentally break it again in the future.

foreverallama:
Good point. I added a write test for an empty dict. It is loaded back as an empty dict, which should be a 1x1 struct with no fields in MATLAB.

test/write.jl
```diff
@@ -193,4 +193,111 @@ nt_read = matread(tmpfile)["nt"]
...
+        read_var = matread(tmpfile)
+
+        @test haskey(read_var, "var1")
+        @test haskey(read_var, "var2")
+        @test isa(read_var["var1"], MatlabOpaque)
+        @test isa(read_var["var2"], MatlabOpaque)
+        @test read_var["var1"] === read_var["var2"]  # same object
+    end
+end
+
```
matthijscox: I don't see tests for the special FileWrapper cases? Is that intentional?

foreverallama: Which one do you mean?

src/MAT_HDF5.jl
```diff
@@ -541,14 +543,19 @@ function m_write(mfile::MatlabHDF5File, parent::HDF5Parent, name::String, c::Abs
...
+function m_write(mfile::MatlabHDF5File, parent::HDF5Parent, name::String, data::AbstractArray{T}, object_decode::UInt32=UInt32(0)) where T
     data = _normalize_arr(data)
     refs = _write_references!(mfile, parent, data)
     # Write the references as the chosen variable
     cset, ctype = create_dataset(parent, name, refs)
     try
         write_dataset(cset, ctype, refs)
-        write_attribute(cset, name_type_attr_matlab, "cell")
+        if object_decode == UInt32(3)
+            write_attribute(cset, object_decode_attr_matlab, object_decode)
```
matthijscox: is this (FileWrapper) case tested?

matthijscox:
Alright, everything looks okay to me. Do you have the power to merge already (since you are a contributor now) or do I have to make you an official maintainer for that?

foreverallama:
I can't merge yet

matthijscox:
I'm only a member of JuliaIO I see, so cannot add you as maintainer unfortunately. I'll just have to merge PRs I suppose

foreverallama:
No issues, you can also close #214

PR #210, opened 2025-11-25
Strip null characters from variable name in `getvarnames`

drbergman:
This udpates `getvarnames` to match the functionality of `read_matrix` using `strip` to deal with the possibility of terminal (or beginning) null characters.

Note: `getvarnames` is only used when 1) checking if a `Matlabv4File` has a specific key, 2) getting the keys of such a file OR 3) `read`ing a specific variable from such a file.

### DIFF
--- a/src/MAT_v4.jl
+++ b/src/MAT_v4.jl
@@ -187,7 +187,7 @@ function getvarnames(matfile::Matlabv4File)
             M, O, P, T, mrows, ncols, imagf, namlen = read_header(matfile.ios, matfile.swap_bytes)
             f = matfile.ios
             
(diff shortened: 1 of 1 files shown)


matthijscox:
Looks fine. I assume you encountered a file with an error/failure? Do we want to add a small unit test to avoid this failure in the future?
If you have no time, I'm okay with merging without testing. Seems like a small risk of future mistakes.

matthijscox:
> Looks fine. I assume you encountered a file with an error/failure? Do we want to add a small unit test to avoid this failure in the future? If you have no time, I'm okay with merging without testing. Seems like a small risk of future mistakes.

Whatever. merging this now so I can include it as a fix in v0.11.1

drbergman:
FWIW: I did have two v4 .mat files, one w/ and one w/o the trailing null character.

Issue #215, opened 2025-12-01
Variables of type duration, datetime, categorical, and table created in MATLAB 2024b cannot be read.


ZhuWenhao-88:
I created some variables of types such as duration, datetime, categorical, and table using MATLAB and saved them as .mat files. The files were created and saved using MATLAB 2020b. They can be read normally using the latest master branch's MAT.jl, but MATLAB 2024b and maltab 2025b will report errors (the problem with string types in the master branch has been fixed).

<img width="1741" height="371" alt="Image" src="https://github.com/user-attachments/assets/3e99865e-f0ee-447f-b323-c7a15d5e295a" />

<img width="1382" height="758" alt="Image" src="https://github.com/user-attachments/assets/810b6e34-37b6-4390-9cee-5a6cd68dcd31" />

<img width="1403" height="768" alt="Image" src="https://github.com/user-attachments/assets/e95b144e-281a-4143-a3f6-9a66be8fffdf" />

[create_data.txt](https://github.com/user-attachments/files/23844248/create_data.txt)
[julia_test_data_2020b.txt](https://github.com/user-attachments/files/23844246/julia_test_data_2020b.txt)
[julia_test_data_2024b.txt](https://github.com/user-attachments/files/23844249/julia_test_data_2024b.txt)
[julia_test_data_2025b.txt](https://github.com/user-attachments/files/23844247/julia_test_data_2025b.txt)

MegaShark1911:
the datetime variable inside 2020b and 2024b are using different format or precision, in 2020b, the datenum field is saved in 7 bytes, while 2024b using 8 + 3 bytes. that's pretty strange.

1. why the precision is different between 2020b and 2024b,  @ZhuWenhao-88  are you using the same command: `dt_scalar = datetime('now');` to generate the dt_scalar?

2.  what's the purpose of  matlab 2024 using the extra 3 bytes , in the array flags, the 2020b is 06 while 2024b is 06 08.
 
3. for the MAT.jl, I think the current code logic didn't forsee the two double data element within a 1*1 matrix, so it interpet both of the data element caused the error. Again, this matlab newer design is weired!!!

ZhuWenhao-88:
Yes, I used the same code, and the 'm' code was created in the create_data file.

foreverallama:
The issue is only with datetime in the newer versions, or other types as well?

MegaShark1911:
I think only with newer version datetime, didn't check the other mcos type for details

[...]

Issue #212, opened 2025-11-28
Unable to read MATLAB 2024b .mat file containing string type


ZhuWenhao-88:
I am using MAT.jl v1.11. When I use MATLAB 2020b to save a string as a .mat file, it can be read normally, but when I use MATLAB 2024b, it cannot be read normally.
<img width="1387" height="646" alt="Image" src="https://github.com/user-attachments/assets/fc144aab-d56f-4a09-b865-3389b9e4ecd7" />

MegaShark1911:
The reason might be the difference between mat file V6 and V7, mat file V7 use Unicode character encoding. 
Can you provide these two mat files for further investigation.

ZhuWenhao-88:
It seems unable to upload .mat files; these two are the corresponding files. In fact, both should be version 7.3; even after adding -v7.3, the saved .mat files differ.
[str2020.txt](https://github.com/user-attachments/files/23813791/str2020.txt)
[str2024.txt](https://github.com/user-attachments/files/23813792/str2024.txt)

MegaShark1911:
The MCOS structure is different between mat files saved by MATLAB2020 and MATLAB2024, I guess @foreverallama may only do the reverse engineering for the first one, need to modify the codes to support the later MCOS structure.
Main difference is in Cell[-3]-Cell[-1].


I have uploaded the manually uncompressed mat files for quick check.

[str2020-uncompressed.txt](https://github.com/user-attachments/files/23814593/str2020-uncompressed.txt)
[str2024-uncompressed.txt](https://github.com/user-attachments/files/23814594/str2024-uncompressed.txt)

foreverallama:
It's actually not an issue with the string or MAT-file, but an issue with the load logic itself. Looks like MATLAB has slightly tweaked the internals in 2024b. It's a very small fix, will push it soon.

PR #216, opened 2025-12-02
Edit: Fix loading datetime with complex valued data in MAT-files

foreverallama:
This is a small fix towards #212 and #215 (builds on #213 ) where it looks like MATLAB has made a few subtle changes in newer versions:

1. MATLAB datetime appears to store some sort of time precision in a complex number. I'm not sure what it is exactly, and until it can be figured out this a temporary fix to prevent errors due to float remainders when converting to Julia DateTime.

2. Newer MATLAB versions seem to maintain a separate object ID counter for some built-in types like `string` and maybe even `timetable` which are serialized using a `saveobj` method. This means that these objects no more contain a `dep_id` which was used to deserialize dynamic properties. The fix takes care of this case. This update shouldn't be a problem for writing as MATLAB should be maintaining backward compatibility.


matthijscox:
No unit tests needed you think?

foreverallama:
Not sure. Do we just keep adding MAT-files for different MATLAB versions? Will be good to make sure we don't break anything backwards, but I don't have access to the latest MATLAB version to generate test cases.

Any case, don't think this is ready to merge yet as the intended behaviour needs to be better understood for both issues addressed in this PR.

matthijscox:
You know what, I'll install 2024b for this reason alone :) Will probably take a few hours.

I do agree we should be careful with polluting the package with test files. But perhaps this is important enough to add one small 2024 test file?

matthijscox:
I've created the request R2024b .mat file for testing, let me know if you need more.

I'm also informed that we could potentially get free github cloud licenses to run multiple MATLAB versions in this open source project, see [MATLAB actions / setup MATLAB](https://github.com/matlab-actions/setup-matlab). For example to automatically generate the test files on the latest MATLAB version. If we keep running into similar issues that might be worth the effort.

foreverallama:
That's very interesting, thanks! Thanks for the test file as well, I'll study it later. I've been looking at some other MAT-files as well, and so far I've figured out the following:

The FileWrapper object contains a version indicator somewhere. However, this only seems to be a major version. I came across a few examples where the FileWrapper metadata format had slightly changed within the same major version. As far as I know, there isn't any indicator of a minor version anywhere.

I believe we've covered both major and minor versions up to 4.x. There's a few metadata regions whose purpose we don't know about yet, but which MATLAB has reserved for future updates. So we won't need to worry about any significant changes to the MCOS codebase, Mathworks seems to have planned this out for a long time now.

We could add some sanity checks to trigger warnings when the reserved regions contain data, this would also help us keep up to date with MATLAB updates.

matthijscox:
Excellent! Let me know whenever you are ready to merge this PR

foreverallama:
Turns out it wasn't actually an issue with the new MAT-file, but a tiny bug that passed through on `MAT_subsys:355` where `get_dynamic_properties()` was supposed to be called using `dep_id` instead of `object_id`. The minor update in the new MAT-file is a non-issue, for dynamic properties particularly.

I don't think we need a separate test for this. It would be better to hardcode a test using a dummy `subsys` instance instead of relying on MAT-files. I'll try and get to it some other time. 

However, I think we should split the datetime fix in `MAT_types` in a different PR. I just checked again, and using `datetime('now')` in MATLAB2023b saves data as a complex number as well, so this is probably not a version related issue. The fix here works, but it might a better idea to record it separately as it might result in loss of precision. We could merge a new test for datetime then.

matthijscox:
Sure, go ahead, and feel free to open or update issues if you fear you might forget your ideas.

I assume you'll remove the datetime fix from this PR then?

foreverallama:
I changed this to the datetime PR. Edited title to reflect it.

Summary of changes:
- Updated `struct_table_datetime.mat` with a new field `testDatetimeComplex` which should contain complex valued datetime data generated using `datetime('now')` in MATLAB 2023b. The update to the MAT-file is for both v7 and v7.3.
- The fix resolves errors due to sub-ms precision encoded in the complex part. However, this means that we lose sub-ms precision in this case. Exactly how MATLAB is encoding precision is unclear as the real part also appears to contain sub-ms precision, so this is a temporary fix

Issue #180, opened 2023-01-26
Invalid parsing of unicode in strings


exaexa:
Hello,

we spotted a several .mat files that use unicode in strings, such as the one here:

https://github.com/SysBioChalmers/Fruitfly-GEM/blob/main/model/Fruitfly-GEM.mat

Strings in these files contain non-ascii characters such as α and β, unfortunately produce something like:
```
 "\x03-Est1"
 "\x03-Est10"
 "\x03-Est2"
 "\x03-Est3"
 "\x03-Est4"
 "\x03-Est5"
 "\x03-Est6"
```
(this is in the `"genes"` sub-array).

Is there any way to specify the decoding of strings or any other way to fix this?

Thank you!

-mk

cc: @htpusa @laurentheirendt

matthijscox:
Note this is purely a MAT v7 thing for certain characters like α and β. When you save with `save('test.mat', 'str', '-v7.3')` it works fine. When you try to save with `save('test.mat', 'str', '-v6')` you get an error in MATLAB (`Found characters the default encoding is unable to represent.`)

foreverallama:
The issue is [here](https://github.com/JuliaIO/MAT.jl/blob/7ccc1c3554dfd6a9e7bbc050dcda8119e947ab1d/src/MAT_v5.jl#L295) in `MAT_v5.jl:305`

```julia
if 255 < convert(UInt32, char)
    # Newer versions of MATLAB seem to write some mongrel UTF-8...
    char = String([truncate_to_uint8(chars[i] >> 8), truncate_to_uint8(chars[i])])[1]
end
```
It looks like its dropping the last 8 bits? 

I'm not very familiar with char encodings, but MATLAB [documentation](hworks.com/help/pdf_doc/matlab/matfile_format.pdf) says `dtype=17` contains Unicode UTF-16 Encoded Character Data, so I think the logic here must be tweaked slightly to separately account for `dtype=17`

MegaShark1911:
For Char type variable, the encoding is different between V6 and V7, while V6 use uint16 and V7 use utf-8.
For the String type variable, the actual data content in Cell 3 using utf-16 by default, I think the '1' after Uint64 tag and data length stands for utf-16 marker, default value is '1'.

matthijscox:
This should be fixed with the latest PR, please check MAT.jl 0.12 when possible.

Issue #143, opened 2020-09-16
Cannot write Char array


lstagner:
```julia
julia> using MAT

julia> a = fill(' ',(3,3))
3×3 Array{Char,2}:
 ' '  ' '  ' '
 ' '  ' '  ' '
 ' '  ' '  ' '

julia> matwrite("test.mat",Dict("a"=>a))
ERROR: This is the write function for CompositeKind, but the input doesn't fit
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] m_write(::MAT.MAT_HDF5.MatlabHDF5File, ::HDF5.HDF5Group, ::String, ::Char) at /Users/lstagner/.julia/packages/MAT/XPntV/src/MAT_HDF5.jl:529
 [3] m_write(::MAT.MAT_HDF5.MatlabHDF5File, ::HDF5.HDF5File, ::String, ::Array{Char,2}) at /Users/lstagner/.julia/packages/MAT/XPntV/src/MAT_HDF5.jl:478
 [4] write(::MAT.MAT_HDF5.MatlabHDF5File, ::String, ::Array{Char,2}) at /Users/lstagner/.julia/packages/MAT/XPntV/src/MAT_HDF5.jl:545
 [5] #matwrite#4(::Bool, ::typeof(matwrite), ::String, ::Dict{String,Array{Char,2}}) at /Users/lstagner/.julia/packages/MAT/XPntV/src/MAT.jl:157
 [6] matwrite(::String, ::Dict{String,Array{Char,2}}) at /Users/lstagner/.julia/packages/MAT/XPntV/src/MAT.jl:148
 [7] top-level scope at REPL[36]:1
```

matthijscox:
Doesn't error anymore in MAT v0.11, but we write a cell array now. Probably also not what you expect.

Issue #33, opened 2014-09-24
Char matrix support


egoecho:
Both MATLAB and Julia support charactor matrix, but saving charactor matrix using MAT.jl in Julia will trigger an error. Charactor matrix is still useful when data is large since it is much faster in reading/writing and need fewer memory in MATLAB. Would you please support saving charactor matrix in MAT.jl?

matthijscox:
Should be fixed now; only 12 years later :)

Issue #176, opened 2022-06-26
BoundsError when struct contains 3D char array


TomRottier:
MWE:

In Matlab create struct with 3D char array and save as .mat
```matlab
c1 = ['abc'; 'abc'];
char_array3D = cat(3, c1, c1)
s = struct('c', char_array3D)
save('s.mat', 's')
```

In Julia read .mat file containing 3D char array
```julia
using MAT
d = matread("s.mat")
```

matthijscox:
Note that behavior is different for v7.3 (HDF5 format).

```
save('s.mat', 's', '-v7.3')
```

Now I get UInt16 arrays
```
julia> matread("s.mat")["s"]["c"]
2×3×2 Array{UInt16, 3}:
[:, :, 1] =
 0x0061  0x0062  0x0063
 0x0061  0x0062  0x0063

[:, :, 2] =
 0x0061  0x0062  0x0063
 0x0061  0x0062  0x0063
```

PR #222, opened 2025-12-04
Char array fixes

matthijscox:
I'm trying to address all the char/string array issues. Have been able to fix the v7.3 (HDF5) versions at least. 
The binary version (v6, v7) still eludes me. Opening the PR to hopefully get some feedback/collaboration.

This attempts to address: #180 #143 [#33](https://github.com/JuliaIO/MAT.jl/issues/33)


foreverallama:
I think we can also add #176 here, looks like we currently do not support anything greater than 2D in `v5` at least. I'll try and take a look later today.

This might also be a good time to decide if/how we want separate types for `char` and `string` moving forward to facilitate writing MATLAB char vs string arrays

matthijscox:
Yeah the difference between char array and string is annoying. high-dimensional char arrays are weird anyway. 

Currently a 3-D char array becomes a 2-D String matrix (v7.3), where the 2nd dimension is converted into the strings.

I guess we could decide to only support: `String` (1xN char matrix) and `Vector{String}` (MxN char matrix) and `Array{Char}` otherwise. Or perhaps only `String` and `Array{Char}` for simplicity? (note then we will still support reading Vector{String} via the Matlab string class)

foreverallama:
I don't know user preferences in Julia, so I'll refrain from commenting. Don't think its easy to get any significant opinions from end users as well is there?

I'm seeing the MATLAB string type natively in datatypes like timeseries and timetable. So if we end up supporting write for those datatypes, only then we'll need a clear differentiation between chars and strings. Otherwise its a user preference issue, rather than compatibility.

matthijscox:
Personally I think the N-D char arrays are a niche MATLAB thing, I think very few people care about them, so it shouldn't matter too much. Being able to load them at all is the main priority I would say. However, we can take a moment to choose a good default now.

foreverallama:
> Vector{String} (MxN char matrix) and Array{Char} otherwise

I agree, this sounds good for char then. What are the possibilities for MATLAB strings?

foreverallama:
I think we can proceed after settling on a design choice that preferably does not break compatibility backwards. As I understand, we are considering the following:
- Loading a 1xN char array as `String`
- Loading a MxN char array as `Vector{String}`
- Loading N-D char arrays as `Array{Char}`

This leaves out MATLAB Strings, which are right now also loaded as any of the above (depending on dimensions). It would be ideal to have some way to differentiate the two types (Char vs String), which would make it easier for saving to MAT-files. But I guess we might need to tradeoff somewhere

matthijscox:
I'm assuming this is now addressed in MAT v0.12

PR #234, opened 2026-03-06
RW for char arrays with unicode support

foreverallama:
Extension of #222 that attempts to fix several bugs in handling char arrays. I've attempted a common decode method `decode_char_array` in `MAT_types.jl`. The `MAT_v5` and `MAT_HDF5` readers now just read raw integer data, and then pass that to `decode_char_array` where the data is decoded. Seems to work with N-D arrays without issue.

So far all the test cases for `read` pass. I'll add some more test cases to check this out, particularly with unicode characters. and N-D arrays. Will need to see how to incorporate the same changes in `MAT_v4.jl` as well.

CC @matthijscox


matthijscox:
Looks like a great clean-up!

I'm only slightly worried about performance because I know val-based dispatching can be slower than normal dispatching. See e.g. https://github.com/ztangent/ValSplit.jl

foreverallama:
The reference you linked mentions this:
> Note that dynamic dispatch does not always occur: When there are a small number of values to split on (less than 4, as of Julia 1.6), the Julia compiler automatically generates a switch statement

Here, the `Val` argument has 3 types (UTF-8, UTF-16, UTF-32), so I don't think the dispatching here impacts performance? In any case, an alternative I guess is we could do a manual if-else and do everything inline? Anyways only the `uint16` type array would need the codec.

What do you think?

foreverallama:
I don't think we need to update `MAT_v4`. It doesn't support unicode so we don't need any special handling. The output type/formatting of the v4 reader is consistent with what the other formats.

foreverallama:
Added write support for both Julia string and char types. Added some tests with unicode characters as well. The basic gist is this:
- For string arrays, the max length of a string is added as the 2nd dimension. Additionally, pad other rows with spaces if necessary for uniform width (which is required by MATLAB).
- For char arrays, transcode each character to utf-16 code units and merge along rows. Again, pad rows if necessary.
 
I've added a bunch of tests for both read and write, and they all work with my version of MATLAB (2025b). I think I've handled some edge cases like 1D vectors and stuff as well. 

Honestly there's probably a better way to do this, could use some pointers there. Till then I believe this is a reasonable solution.

matthijscox:
> Here, the Val argument has 3 types (UTF-8, UTF-16, UTF-32), so I don't think the dispatching here impacts performance? In any case, an alternative I guess is we could do a manual if-else and do everything inline? Anyways only the uint16 type array would need the codec. What do you think?

I'm okay with a manual if-else in `_decode_row`, just to be sure. I don't expect us to add many more codecs.

> I don't think we need to update MAT_v4. It doesn't support unicode so we don't need any special handling. The output type/formatting of the v4 reader is consistent with what the other formats.

Alright.

> Honestly there's probably a better way to do this, could use some pointers there. Till then I believe this is a reasonable solution.

At a glance the code looks okay, is there anything in specific you'd like some advice on?

matthijscox:
I'm wondering about the next version bump after this. I think this is a serious behavior change for strings/chars, so I suppose we should go from 0.11 to 0.12.

foreverallama:
Thanks for the review!

> I'm okay with a manual if-else in _decode_row, just to be sure. I don't expect us to add many more codecs.

Great, I'll make a small update for this.

> is there anything in specific you'd like some advice on?

I was just wondering if there's some way to avoid iterating over all elements of the array when encoding

> we should go from 0.11 to 0.12

I agree. I believe there is no issue with backwards compatibility, but it is a major update nonetheless, and could come with its own set of bugs.

foreverallama:
- Updated to a symbol based runtime dispatch of `codec` with `uint16` arrays.
- Fix: Use Endianness based byte swapping for `uint16` arrays using `utf8` encoding
- Simplified if-else flow for determining codecs & reading raw char array data

matthijscox:
Can we consider this PR as finished?

foreverallama:
Yes, I believe the latest changes has improved from where we started and addressed your concerns as well. There is room for optimisation which could be addressed in future PRs.

Immediately, I see a place we could optimise in `_decode_row(row::AbstractVector{UInt16}, codec)`:
- Instead of byte swapping, we could just interchange MSB and LSB based on endianness
- Pre-allocate the new uint8 buffer to max possible size and then shrink at the end before transcoding, instead of `push!` for every byte.

If that sounds good, I'll just refactor that method and I think we're good to go.

The changes here should close #33 #143 #176 and #180

matthijscox:
Sure, go ahead. Also feel free to version bump, so I can register right after merging this PR.

PR #248, opened 2026-09-17
Prepare for HDF5.jl v0.18 (H5DataStore <: AbstractDict)

mkitti:
## Summary

Prepares MAT.jl for HDF5.jl v0.18 (JuliaIO/HDF5.jl#1245, JuliaIO/HDF5.jl#1247), which makes `HDF5.H5DataStore <: AbstractDict{String,Any}`.

`MatlabHDF5File`, `Matlabv5File`, and `Matlabv4File` are all `<: HDF5.H5DataStore`, but MAT.jl defines none of `getindex`/`length`/`get`/`copy`/`empty`/`==`/`hash`/`iterate` on any of them — so they cleanly inherit HDF5.jl's new generic `AbstractDict` fallbacks for all of these. Unlike JLD.jl (see [JuliaIO/JLD.jl#336](https://github.com/JuliaIO/JLD.jl/pull/336)), which has its own non-conforming `iterate` that needed a real fix, MAT.jl needs only the compat bound bumped.

- `Project.toml`: `HDF5 = "0.16, 0.17"` → `"0.16, 0.17, 0.18"`.

## Verification

Ran the full test suite (581 tests) locally, both:
- against the currently released HDF5 0.17 (baseline), and
- against the pending HDF5.jl branch ([`mkitti/HDF5.jl@hdf5-tree-completion`](https://github.com/mkitti/HDF5.jl/tree/hdf5-tree-completion), tip `64ab8913`, which includes #1245's `H5DataStore <: AbstractDict` change).

All 581 tests pass, unmodified, in both cases.

## Why draft

HDF5.jl v0.18 isn't released yet, so the `HDF5 = "...0.18"` compat entry can't resolve against a real registered version and CI here will keep exercising 0.17 for now. Opening this now so MAT.jl is ready to un-draft/merge as soon as HDF5.jl 0.18 ships.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


matthijscox:
Well that's an interesting little tidbit. Matlabv4File and Matlabv5File aren't even HDF5 files, they are custom binary data formats. So now I'm wondering if that's a wrong legacy design choice. I bet I can remove that subtyping entirely, maybe replace it with `Matlabv4File <: IO`.

I'm curious now if any of the HDF5.H5DataStore functionality is even used by MatlabHDF5File.

Anyway, good to know we can bump the compat once HDF5 v0.18 releases, thanks for checking!

